Issue #16, opened 2018-08-01
Improvements to reading measurement results


hellais:
Currently, to avoid some cross build hell that comes with using native npm packages, we use a pure JS SQLite implementation (https://github.com/ooni/probe-desktop/commit/22d792ec4f8511f9a99de26d5fda43635c1ae8ba) that has been transpiled from C code: https://github.com/kripken/sql.js/ (yes it's as sketchy as it sounds).

I am not too pleased with this solution to be honest, and it has been the source of several bugs and workarounds. For example to avoid this bug: https://github.com/kripken/sql.js/issues/113, we import the version of the library that support memory-growth: https://github.com/ooni/probe-desktop/blob/master/main/database.js#L18.

I am thinking that other possible approaches for handling this better could be:

1. Instrument the probe-cli command to read the database and parse the results from stdout (which is a series of JSON documents) and use those to populate the results views (the cons of this approach is that we pay some overhead due to having to start a process every time we want to read results)

2. We implement a HTTP (or websockets) backend inside of probe-cli that gives the measurement results to the electron frontend. I think this requires a bit more work (and can possibly open up security holes), but would also open up the possibility of running the desktop code in a web service environment (think something a-la ooni-wui).

hellais:
> Instrument the probe-cli command to read the database and parse the results from stdout (which is a series of JSON documents) and use those to populate the results views (the cons of this approach is that we pay some overhead due to having to start a process every time we want to read results)

I think the best way forward is to use this approach of spawning a process of ooniprobe and parse the JSON formatted output it already produces as part of `--batch`.
I suspect this would require only modifications to the desktop app and would work with the current `ooni` binary.

If this appears to be too fragile or painful to deal with, we can then try out the other approach.

PR #17, opened 2018-09-11
Feature/new schema

hellais:
This pull request include all the probe-desktop changes that are related to: https://github.com/ooni/probe-cli/pull/16.

As part of this I have also replaced all the logic that reads uses `sql.js` to read from the database and instead we proxy all the measurement listing and reading requests through the `probe-cli` tool. This plan had been brought up in https://github.com/ooni/probe-desktop/issues/16 and I have verified that it doesn't seem to be too painful to do, the only concern I have over this approach is that of scalability, but it seems to be working ok for the test cases I have run into so far.


hellais:
As probe-cli has landed, I am going to land this as well.

As mentioned in the probe-cli: https://github.com/ooni/probe-cli/pull/16#issuecomment-422361395, users prior to this unreleased version should do a hard reset of OONI Probe. This can be done either manually deleting the ooni_home directory or by using the Hard reset button in the about page of OONI Probe desktop.

PR #23, opened 2018-10-08
Allow to build on Linux

### LINKED ISSUE #22: fsevents fails to install as a dependency on Linux
Error log from `yarn install`
> error fsevents@1.2.4: The platform "linux" is incompatible with this module.
> error Found incompatible module

The typical solution seems to be to move `fsevents` under `optionalDependencies` in package.json. e.g [facebook/create-react-app/react-scripts](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/package.json#L76)

sarathms:
Closes #22 

This PR allows development on Linux.

Moved the below set of tasks to support final builds on Linux to #38 

> Adds configuration to generate `.deb` builds for Debian-based Linux distributions.
> 
> **TODO**
> 
> - [ ] Use a high resolution PNG to generate application icons for multiple resolutions (16x16, 32x32... 1024x1024)
> - [ ] Use the same icon as argument to `electron.BrowserWindow` to show app icon in window title bar.
> - [ ] `zip` based builds should extract contents into a subdirectory and not in the containing directory (e.g `${HOME}/Downloads`)


sarathms:
For now, it works fine with an additional manual step of creating `./bin/cert.pem` before running `yarn run download-bin`. The correct solution to this is being discussed in #25

hellais:
Looks good to me.

PR #39, opened 2019-09-16
Remove page title from _document.js

sarathms:
Using `<title>` in `pages/_document.js` leads to unexpected results as per https://github.com/zeit/next.js/blob/master/errors/no-document-title.md

### DIFF
--- a/renderer/pages/_document.js
+++ b/renderer/pages/_document.js
@@ -25,7 +25,6 @@ class CustomDocument extends Document {
     return (
       <html>
         <Head>
-          <title>OONI Probe</title>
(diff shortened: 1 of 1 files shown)


hellais:
lgtm

hellais:
Don't we though want to re-add it to some base layout component?

sarathms:
Electron seems to use the app name as the window title, which is 'OONI Probe' anyway. So it is not empty. Which is why I felt it is OK to conclude this PR. As we find use cases to change the title or find it missing, we can be explicit about it.

PR #41, opened 2019-09-17
Wrap trasnlatable text with FormattedText

### LINKED ISSUE #6: Wrap all translatable text with FormattedMessage
Recently we added support for translations in the desktop app, however there are still some places in which we are not using the translated strings, but instead are using raw strings.

This is about going through the codebase and ensuring all the strings that are to be translated are wrapped with the `<FormattedMessage>` component.

sarathms:
Closes #6

### DIFF
--- a/renderer/components/test-results/ResultRow.js
+++ b/renderer/components/test-results/ResultRow.js
@@ -64,10 +64,14 @@ const SummaryContainer = styled(Flex)`
 const WebsitesSummary = ({anomalyCount, totalCount}) => {
   return <SummaryContainer flexWrap='wrap'>
     <Box width={1}>
-      <Text color={anomalyCount > 0 ? theme.colors.red8 : theme.colors.black}><MdClear /> {anomalyCount} blocked</Text>
+      <Text color={anomalyCount > 0 ? theme.colors.red8 : theme.colors.black}>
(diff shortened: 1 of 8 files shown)


renderer/components/test-results/ResultRow.js
```diff
@@ -64,21 +64,29 @@ const SummaryContainer = styled(Flex)`
...
     <Box width={1}>
-      <Text color={anomalyCount > 0 ? theme.colors.red8 : theme.colors.black}><MdClear /> {anomalyCount} blocked</Text>
+      <Text color={anomalyCount > 0 ? theme.colors.red8 : theme.colors.black}>
+        <MdClear /><FormattedMessage id='TestResults.Overview.InstantMessaging.Blocked.Plural' values={{ Count: anomalyCount }}/>
+      </Text>
     </Box>
     <Box width={1}>
-      <Text><MdDone /> {totalCount} tested</Text>
+      <Text>
+        <MdDone /> <FormattedMessage id='TestResults.Overview.InstantMessaging.Available.Plural' values={{ Count: totalCount }} />
```
sarathms: @hellais On the mobile app, this part shows the number of available IMs instead of `totalCount` like in Website summary. Should we switch this also to show `totalCount - anomalyCount` here?

hellais: Yeah let's align it to what we are doing on mobile.

PR #42, opened 2019-09-18
Remove deprecated prefetch attribute from Link

sarathms:
Next.js 9 automatically prefetches components as they appear in viewport. Refer: https://err.sh/zeit/next.js/prefetch-true-deprecated


hellais:
sgtm

PR #49, opened 2019-10-08
Implement measurement screen for Web Connectivity tests

sarathms:
This implements the details view for WebConnectivity tests. I used the render props pattern 

**Reachable**
Screenshot from 2019-10-08 16-08-28

**Anomaly**
Screenshot from 2019-10-08 16-10-03


sarathms:
I used the [render props pattern](https://reactjs.org/docs/render-props.html) to pass parameters up the render tree to parent components. More specifically, the test-specific component for details (in this case `components/nettests/websites/WebConnectivity.js`) sends pieces of information to its parent component `MeasurementContainer` to construct the Hero component, which needs to know what `WebConnectivity.js` wants to convey.

The traditional way to do this would be do let the test-specific components render the entire screen (Hero + Details) in their own render function. This would obviously result in duplicating a pile of code across each test screen.

The test detail component can also build the entire content for `<Hero>` if it choose to, and pass it in the `hero` of the render function, which would then make `<MeasurementContainer>` use that one instead of the templated one defined in `<StickyHero>`. This can be useful when implementing `<Dash>` and `<NDT>` down the line.

renderer/components/FormattedMarkdownMessage.js
```diff
@@ -39,8 +39,8 @@ class FormattedMarkdownMessage extends React.Component {
   render() {
     const { formatMessage } = this.context.intl
     const { id, defaultMessage, values, description } = this.props
-    const messageDescriptor = { id, defaultMessage, values, description }
-    const message = formatMessage(messageDescriptor)
+    const messageDescriptor = { id, defaultMessage, description }
+    const message = formatMessage(messageDescriptor, values)
```
sarathms: The correct syntax for `formatMessage` [as per docs](https://github.com/formatjs/react-intl/blob/master/docs/API.md#formatmessage).

data/lang-en.csv
```diff
@@ -380,4 +380,8 @@ CategoryCode.GOVT.Description,"Government-run websites, including military sites
 CategoryCode.COMM.Description,Sites of commercial services and products.,
 CategoryCode.CTRL.Description,Benign or innocuous content used as a control.,
 CategoryCode.IGO.Description,"Sites of intergovernmental organizations, such as the United Nations.",
-CategoryCode.MISC.Description,Sites that haven't been categorized yet.,
\ No newline at end of file
+CategoryCode.MISC.Description,Sites that haven't been categorized yet.,
+TestResults.Details.Websites.LikelyBlocked.Content.Paragraph,"{{WebsiteURL}} is likely blocked by means of {{BlockingReason}}.
+
+Note: False positives can occur. Learn more [here](https://ooni.io/nettest/web-connectivity/).",
+TestResults.Details.Websites.Reachable.Content.Paragraph,"{{WebsiteURL}} is accessible.",
```
bassosimone: How does a CSV deal with newlines? I am surprised the `Note: False positive [...]` is parsed as `TestResults.Details.Websites.LikelyBlocked.Content.Paragraph`.

On a related note, is it correct to say that the JSON derives from the CSV? What is the process?

bassosimone:
> One interesting thing to note about render props is that you can implement most higher-order components (HOC) using a regular component with a render prop.

Seems potentially useful for ReactNative!

data/lang-en.csv
sarathms: As long as new lines are enclosed in `""` it will be parsed as a single string until the next `,` is found.

Yes, the `lang/en.json` file is derived from the CSV file. This is done by `scripts/update-translations.js`. The translation workflow is currently in a mess for `probe-desktop`. It currently uses an older version of a source CSV that was used by `probe-mobile` which has now moved onto what is available in `ooni/translations`. So there is a bunch of work to be done here to sync up the keys. Will open an issue for this.

PR #51, opened 2019-10-10
Upgrade to react-intl@3.2.2

sarathms:
This allows us to use the 'useIntl' hook which is a
simpler way to invoke the imperative API like
formatMessage to translate JS strings not components.


renderer/components/withIntl.js
```diff
@@ -14,12 +14,6 @@ const getLocale = () => {
   return navigatorLang.split('-')[0]
 }
 
-if (typeof window !== 'undefined' && window.ReactIntlLocaleData) {
-  Object.keys(window.ReactIntlLocaleData).forEach((lang) => {
-    addLocaleData(window.ReactIntlLocaleData[lang])
-  })
-}
-
```
sarathms: This is removed because `react-intl` no longer provides locale data or the method `addLocaleData` to add such locale data. It is expected that modern browsers already implement them and recommended polyfills are included. However, in our case, we render in a controlled environment (unlike explorer) and can afford to not have the polyfills as we remove this functionality.

sarathms: context: "Locale Data" refers to date formats, number formats that are specific to locales.

renderer/components/FormattedMarkdownMessage.js
```diff
@@ -35,23 +35,16 @@ const remarkReactComponents = {
...
-    const messageDescriptor = { id, defaultMessage, description }
-    const message = formatMessage(messageDescriptor, values)
-    // When the message is the ID means it's not defined and we just emit not
-    // element
-    if (message === id) {
-      return null
-    }
-    return remark().use(reactRenderer, {remarkReactComponents}).processSync(message).contents
+const FormattedMarkdownMessage = ({ id, defaultMessage, values, description }) => {
+  const { formatMessage } = useIntl()
```
sarathms: Using the hook instead of the context. Doesn't make a huge difference, just my new-found fascination for hooks :-)

renderer/components/withIntl.js
```diff
@@ -14,19 +14,11 @@ const getLocale = () => {
...
 if (typeof window !== 'undefined' && window.OONITranslations) {
   supportedMessages = window.OONITranslations
 }
 
 const withIntl = (Page) => {
-  const IntlPage = injectIntl(Page)
```
sarathms: There was no need to call the HoC here. Looks like left over from old template for integrating `react-intl` in `next.js`

bassosimone: What's HoC?

package.json
```diff
@@ -108,7 +108,7 @@
     "react-animations": "^0.1.0",
     "react-dom": "16.9.0",
     "react-icons": "3.7.0",
-    "react-intl": "^2.4.0",
+    "react-intl": "3.3.2",
```
bassosimone: you are also dropping the `^` here, why?

renderer/components/FormattedMarkdownMessage.js
```diff
@@ -2,7 +2,7 @@
 import React from 'react'
 
 import PropTypes from 'prop-types'
-
+import { useIntl } from 'react-intl'
```
bassosimone: what code style do you use for grouping imports?

renderer/components/FormattedMarkdownMessage.js
bassosimone: what is the difference?

renderer/components/FormattedMarkdownMessage.js
```diff
@@ -35,23 +35,16 @@ const remarkReactComponents = {
...
-    if (message === id) {
-      return null
-    }
-    return remark().use(reactRenderer, {remarkReactComponents}).processSync(message).contents
+const FormattedMarkdownMessage = ({ id, defaultMessage, values, description }) => {
+  const { formatMessage } = useIntl()
+  const messageDescriptor = { id, defaultMessage, description }
+  const message = formatMessage(messageDescriptor, values)
+  // When the message is the ID means it's not defined and we just emit not
+  // element
```
bassosimone: what does "not element" mean?

package.json
hellais: This means that the dependency is specified more strictly which I guess is not a bad idea.

renderer/components/FormattedMarkdownMessage.js
hellais: It means that instead of defining a class on which you access the `this.context` part of the instance in the render() method you define a pure function on which you call useIntl() which returns a hook.

This hook allows you to make changes to the state of the component without needing to define it as a class.

Effectively the result of this it makes the amount of boilerplate required to do some set of common tasks easier. This feature exists since react 16.8, see: https://reactjs.org/docs/hooks-intro.html.

renderer/components/FormattedMarkdownMessage.js
hellais: I suspect this is a typo and can be rephrased as:

> When the message is the ID means it's not defined, so we don't emit an element

I do wonder, though if this is desired behaviour. It may be useful in these cases to emit a special element which puts a placeholder in the DOM so that it's possible to see what are the strings that still need to be populated with items.

@sarathms what do you think?

renderer/components/withIntl.js
hellais: Higher order Component, it's basically a fancy way of saying that you wrap one component with another one, in this case you are using the `injectIntl` HoC to wrap `Page`.

I believe what this thing was doing is add some instance attributes to the Page to make it work with translations.

I suppose this is no longer needed because in more modern versions of react you can make use of Contexts and therefore it's not needed to wrap stuff in HoCs.

renderer/components/withIntl.js
```diff
@@ -39,7 +31,7 @@ const withIntl = (Page) => {
 
       return (
         <IntlProvider locale={locale} messages={messages} initialNow={now}>
```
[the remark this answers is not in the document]
sarathms: More specifically, `injectIntl` was used to inject the context data into the component being wrapped by the HoC. `<IntlProvider>` is _and_ was necessary to produce the 'context' that is consumed by `injectIntl` HoC earlier or `useIntl` hook now.

renderer/components/FormattedMarkdownMessage.js
sarathms: Good catch. I overlooked it. I don't think it is a good idea to return `null`. Best would be to just render the undefined key as-is which then aligns with the default behavior of `<FormattedMessage>`. Typically, our keys are pretty long and will most likely stick out enough to be noticeable.

Moreover, we can adapt the script we wrote for explorer which detects keys defined but not translated and stale keys in the csv.

package.json
sarathms: `yarn` adds the ^ by default when you don't specify a version when installing. I had specified `3.3.2` in this case.

I am not sure if it is a good idea always to keep it this strict. e.g for well maintained packages, we might miss out on security updates or bugfix patches unless we actively commit those version upgrades in our repo. If we say `react: ~16.9.0` will ensure that we need not update the dependency until we want to switch to `v17`

sarathms: I found this very useful for a quick reference: https://semver.npmjs.com/

PR #52, opened 2019-10-11
Use show command

sarathms:
Integrates the `ooniprobe show` command to fetch raw measurement JSON for a given `measurement_id`.

### DIFF
--- a/main/actions.js
+++ b/main/actions.js
@@ -85,8 +85,38 @@ const listResults = () => {
   })
 }
 
+const showMeasurement = (msmtID) => {
+  const ooni = new Ooniprobe()
+  let measurement = {}
+
+  return new Promise((resolve, reject) => {
+    ooni.on('data', (data) => {
+      if (data.level == 'error') {
+        reject(data.message)
+      }
+
+      switch(data.fields.type) {
+      case 'measurement_json':
+        measurement = data.fields.measurement_json
+        break
+      default:
+        debug('extra data.fields', data.fields)
+      }
+    })
+
+    ooni
+      .call(['show', msmtID])
+      .then(() => {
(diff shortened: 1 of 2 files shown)


main/actions.js
```diff
@@ -85,8 +85,38 @@ const listResults = () => {
   })
 }
 
+const showMeasurement = (msmtID) => {
+  const ooni = new Ooniprobe()
+  let measurement = {}
+
+  return new Promise((resolve, reject) => {
+    ooni.on('data', (data) => {
+      if (data.level == 'error') {
```
bassosimone: shouldn't this be `===`?

main/actions.js
```diff
@@ -85,8 +85,38 @@ const listResults = () => {
...
+        reject(data.message)
```
bassosimone: How does `reject` work? Does it `throw`? Shouldn't you `return` here?

main/actions.js
sarathms: Yes. Fixed.

main/actions.js
sarathms: `reject` doesn't 'throw'. It just passes control to the `catch` block of the promise handler. Not `catch`ing will throw an `Unhandled rejection exception` though.

And you are right. It is better to `return` here. Although not necessarily a problem.

hellais: It's probably then an opportunity to apply this pattern to the other commands as well which don't.

As you said it should not however be a problem as you can only make a `Promise` reach it's final state (either `reject` or `resolve`) and any further call to `reject` or `resolve` shall result in a noop.

It's still probably a good idea to not waste time on those extra operations below in case of error and it's wise to fail early.

sarathms: Done

PR #53, opened 2019-10-11
Show middlebox measurement details

sarathms:
Implements views for the two middlebox tests.

Screenshot from 2019-10-11 16-23-39

Screenshot from 2019-10-11 16-24-49

### DIFF
--- a/renderer/components/measurement/MeasurementContainer.js
+++ b/renderer/components/measurement/MeasurementContainer.js
@@ -21,17 +21,18 @@ import { StickyContainer, Sticky } from 'react-sticky'
 import { FacebookMessengerDetails } from '../nettests/im/facebook-messenger'
 import { WebConnectivity } from '../nettests/websites/WebConnectivity'
 import { HttpHeaderFieldManipulation } from '../nettests/middleboxes/HttpHeaderFieldManipulation'
+import { HttpInvalidRequestLine } from '../nettests/middleboxes/HttpInvalidRequestLine'
 import FullHeightFlex from '../FullHeightFlex'
 
 // TODO: (sarathms) Add rest of the implementations when ready
 const detailsMap = {
   web_connectivity: WebConnectivity,
   http_header_field_manipulation: HttpHeaderFieldManipulation,
+  http_invalid_request_line: HttpInvalidRequestLine,
   // dash: DashDetails,
   // ndt: NdtDetails,
   // whatsapp: WhatsAppDetails,
   // telegram: TelegramDetails,
-  // http_invalid_request_line: HttpInvalidRequestLine,
   // vanilla_tor: VanillaTorDetails,
   facebook_messenger: FacebookMessengerDetails
 }
--- a/renderer/components/measurement/MeasurementContainer.js
+++ b/renderer/components/measurement/MeasurementContainer.js
@@ -20,11 +20,19 @@ import { StickyContainer, Sticky } from 'react-sticky'
 
 import { FacebookMessengerDetails } from '../nettests/im/facebook-messenger'
 import { WebConnectivity } from '../nettests/websites/WebConnectivity'
+import { HttpHeaderFieldManipulation } from '../nettests/middleboxes/HttpHeaderFieldManipulation'
 import FullHeightFlex from '../FullHeightFlex'
 
 // TODO: (sarathms) Add rest of the implementations when ready
 const detailsMap = {
   web_connectivity: WebConnectivity,
+  http_header_field_manipulation: HttpHeaderFieldManipulation,
+  // dash: DashDetails,
+  // ndt: NdtDetails,
+  // whatsapp: WhatsAppDetails,
+  // telegram: TelegramDetails,
+  // http_invalid_request_line: HttpInvalidRequestLine,
+  // vanilla_tor: VanillaTorDetails,
   facebook_messenger: FacebookMessengerDetails
 }
 
--- a/renderer/components/measurement/MeasurementContainer.js
+++ b/renderer/components/measurement/MeasurementContainer.js
@@ -20,6 +20,7 @@ import { StickyContainer, Sticky } from 'react-sticky'
 
 import { FacebookMessengerDetails } from '../nettests/im/facebook-messenger'
 import { WebConnectivity } from '../nettests/websites/WebConnectivity'
+import FullHeightFlex from '../FullHeightFlex'
 
 // TODO: (sarathms) Add rest of the implementations when ready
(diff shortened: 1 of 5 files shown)


sarathms:
This is how we show it on the mobile. However, the early desktop mockups suggest we wanted to also show details like the side-by-side comparison of what headers were sent and received etc. Do we want to do this now? @hellais

bassosimone:
@sarathms I believe we should create an issue for enhancing the view as you describe but we should make sure other higher priority endeavours are addressed first. (I.e. "sprint backlog"?)

renderer/components/nettests/middleboxes/HttpHeaderFieldManipulation.js
```diff
@@ -0,0 +1,56 @@
...
+      })}
+    </div>
+  )
+}
+
+HttpHeaderFieldManipulation.propTypes = {
+  measurement: PropTypes.object,
+  isAnomaly: PropTypes.bool,
+  render: PropTypes.func
+}
```
bassosimone: This is a rather basic question, which however it's probably worth asking. `HttpHeaderFieldManipulation` is a function that maps `isAnomaly, render` to a rendered div. Then you are adding a `.propTypes` field to the function, which is fine, because a function is a normal JavaScript object, except that it's callable.

Upon closer inspection, this seems the declaration of a "prototype". You are telling React (?), that `measurement` is an object, `isAnomaly` is a boolean, and `render` is a function.

Why is `measurement` required? Where is it used?

renderer/components/nettests/middleboxes/HttpHeaderFieldManipulation.js
```diff
@@ -0,0 +1,56 @@
...
+
+export { HttpHeaderFieldManipulation }
+
+export default {
+  name: <FormattedMessage id='Test.HTTPHeaderFieldManipulation.Fullname' />
+}
```
bassosimone: Does this mean that when we import this module, if we don't specify a name to import from the module, we get by default this specific formatted message?

renderer/components/nettests/websites/WebConnectivity.js
```diff
@@ -29,32 +29,30 @@ const WebConnectivity = ({measurement, isAnomaly, render}) => {
...
+            <FormattedMessage
+              id='TestResults.Details.Websites.Reachable.Content.Paragraph'
+              values={{
+                WebsiteURL: url
+              }}
+            />
+          </Box>
+        )}
+      </Flex>
+    </Box>
```
bassosimone: You moved the `FullHeightFlex` up one level (just to check I'm understanding the PR flow)

renderer/components/measurement/MeasurementContainer.js
```diff
@@ -20,10 +20,20 @@ import { StickyContainer, Sticky } from 'react-sticky'
...
 const detailsMap = {
   web_connectivity: WebConnectivity,
+  http_header_field_manipulation: HttpHeaderFieldManipulation,
+  http_invalid_request_line: HttpInvalidRequestLine,
+  // dash: DashDetails,
+  // ndt: NdtDetails,
+  // whatsapp: WhatsAppDetails,
+  // telegram: TelegramDetails,
+  // vanilla_tor: VanillaTorDetails,
   facebook_messenger: FacebookMessengerDetails
```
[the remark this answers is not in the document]
sarathms: TL;DR - It is temporary. Will become consistent very soon.

`FacebookMesssengerDetails` was an early implementation and uses that name. I also started that way, but then switched to dropping the suffix as I finished the first test details component. I have been renaming the placeholders as I implement them and uncomment them. `FacebookMessengerDetails` will also be refactored when I re-implement in the new model.

Did you mean that it is better to have the suffix than not, for readability purposes?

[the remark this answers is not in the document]
bassosimone: I have no strong opinion, and I'll defer to your instincts. I'm mostly concerned about keeping the codebase and the coding style consistent.

renderer/components/nettests/index.js
```diff
@@ -98,8 +100,8 @@ export const testGroups = {
 export const tests = {
   // TODO: (sarathms) Replace each one with their own implementations when ready
   web_connectivity,
-  http_header_field_manipulation: web_connectivity,
-  http_invalid_request_line: web_connectivity,
+  http_header_field_manipulation,
+  http_invalid_request_line: http_invalid_request_line,
```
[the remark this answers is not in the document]
sarathms: You are right. I updated this.

hellais:
> Screenshot from 2019-10-11 16-23-39


The green on the side looks a bit off to me in these screenshots as if they are two slightly different green colors.

renderer/components/nettests/middleboxes/HttpHeaderFieldManipulation.js
sarathms: Yes. This allows us to [use this pattern](https://github.com/ooni/probe-desktop/blob/ux/middlebox-measurements/renderer/components/measurement/MeasurementContainer.js#L21-L24).

renderer/components/nettests/index.js
```diff
@@ -98,8 +100,9 @@ export const testGroups = {
 export const tests = {
   // TODO: (sarathms) Replace each one with their own implementations when ready
   web_connectivity,
-  http_header_field_manipulation: web_connectivity,
-  http_invalid_request_line: web_connectivity,
+  http_header_field_manipulation,
+  http_invalid_request_line,
+  facebook_messenger: web_connectivity,
```
hellais: Is it desired that this points to `web_connectivity`, are you doing that because you don't have an implementation of `facebook_messenger` yet so you are using web_connectity as a placeholder?

renderer/components/nettests/websites/WebConnectivity.js
sarathms: Exactly. Inspired by you, I started using more detailed commit messages like in this case a28888cb4a1ce23f460185ac5efb379c67386a1f.

renderer/components/nettests/index.js
sarathms: Yes. It is a placeholder. It doesn't make a lot of difference, because that implementation isn't stable anymore anyway. This assignment was only to avoid errors at the `MeasurementContainer` level.

sarathms:
Yes. I had noticed it too. I tried debugging, but couldn't figure out immediately what is causing it. You can see that in the anomaly hero too. Will open an issue for this.

PR #62, opened 2019-10-23
Update onboarding UI

### LINKED ISSUE #8: Style the onboarding based on the mockups
The current onboarding was implemented without mockups and as such is not aligned to what it should really look like.

This ticket is about aligning the onboarding views to the mockups.

See: https://xd.adobe.com/spec/aa4af2a6-9501-41ed-53ed-d7dd933947b8-dc7b/

### LINKED ISSUE #19: Add animations on wrong and correct answers in onboarding
Currently the onboarding doesn't have the animations for the correct and wrong answers.

sarathms:
Needs work on:
- [ ] Background images for the onboarding steps
- [x] Transitioning out of the animations

Fixes #8 
Fixes #19


sarathms:
Animations work, but the `complete` event doesn't fire for some reason and doesn't let us unmount the animation and show the next question. [This part](https://github.com/ooni/probe-desktop/pull/62/files#diff-51c5bb6e90b2d6ecdaea6ca065fec954R173-R175) is relevant.

renderer/components/onboard/Sections.js
```diff
@@ -139,18 +154,43 @@ const QuizQuestion = ({qNum, question, onTrue, onFalse}) => (
...
+      options={{
+        loop: false,
+        autoplay: true,
+        animationData: animationData,
+        rendererSettings: {
+          preserveAspectRatio: 'xMidYMid slice'
+        },
+        eventListeners: [{
+          eventName: 'complete',
+          callback: () => this.onAnimatonComplete(),
```
hellais: Could it be that this is not firing due to the fact that inside of this scope `this` is referring to the `{}` object instead of the `Animation` class instance?

Maybe you want to copy `onAnimationComplete` onto the stack before passing it as a callback to this?

sarathms: Ah. I don't think that is the problem. I missed moving out the callback along with the `<Lottie ... />`. But it didn't work even when it was inside. I have reverted to that version now.

sarathms: Turns out `eventListeners` is a prop to `<Lottie>` :man_facepalming:

sarathms:
This is functional now. We just need to wire up the right background images.

lorenzoPrimi:
Animations:
https://github.com/ooni/probe-ios/tree/master/ooniprobe/Resources/anim
https://github.com/ooni/probe-android/tree/master/app/src/main/assets/anim

sarathms:
@hellais 
1. Can you check if you are able to drag with the topbar on onboarding screen?
2. With the new animations, do you get the same sense of lag? I didn't find any difference by adding a `speed={2}` prop to `<Lottie>` in [here](https://github.com/ooni/probe-desktop/pull/62/files#diff-6a2c49729d722ca81608dfb57ac61b19R124).

hellais:
> Can you check if you are able to drag with the topbar on onboarding screen?

Yes the dragging now works great!

> With the new animations, do you get the same sense of lag? I didn't find any difference by adding a speed={2} prop to <Lottie> in here.

It may have been an issue with the other animations you were using. I don't see anymore lag now.

sarathms:
I made the quiz modal (and animations) in the wider aspect ratio before.

PR #64, opened 2019-10-30
Improvements to the result listing

hellais:
This branch adds support for hiding months in the listing that don't have any measurements for them.

### DIFF
--- a/renderer/components/test-results/TestResultsContainer.js
+++ b/renderer/components/test-results/TestResultsContainer.js
@@ -123,6 +123,7 @@ const groupRowsByMonth = (rows) => {
     byMonth[month].push(row)
   })
   return Object.keys(byMonth)
+    // Filter out months with no measurements in them to avoid showing empty months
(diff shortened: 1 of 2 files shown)


renderer/components/test-results/TestResultsContainer.js
```diff
@@ -107,21 +107,26 @@ const groupRowsByMonth = (rows) => {
...
+
   // XXX I think there is a better way to do this without so many list
   // reversals.
   rows.reverse().map(row => {
     const month = moment(row.start_time).format('YYYY-MM-01')
     byMonth[month].push(row)
   })
-  return Object.keys(byMonth).sort().reverse().map(key => [key, byMonth[key]])
+  return Object.keys(byMonth)
+    .filter(month => byMonth[month].length > 0 )
```
sarathms: Is this to avoid the empty section in the result list? Might be useful to add the reason here.

hellais: Done.

PR #70, opened 2019-10-31
Upgrade electron-builder

sarathms:
Upgraded.

I still can't get the build to run properly. It opens up the main and about windows that are blank.


hellais:
I don't think this is necessary.

This fix https://github.com/ooni/probe-desktop/pull/73 does it.

We should eventually upgrade electron-builder, but probably not right before a release.

PR #75, opened 2019-10-31
Finish measurement pages

### LINKED ISSUE #66: Finish the measurement pages
Currently several bits are missing from the measurement page:
* [x] Methodology link should point to the nettest pages on our website and use a styled button
* [x] Raw Data should open the raw data and use a styled button
* [x] Explorer URL should point to the explorer address for the measurement (the format of this is `https://explorer.ooni.org/$REPORT_ID?input=$INPUT`) and it should be omitted if the result is not uploaded
* [x] Remove the view log button as I don't think we have support for logs

sarathms:
Fixes #66


hellais:
LGTM!

Merging into the stable-release branch.

Issue #63, opened 2019-10-30
Card descriptions overflow when clicking on the ? icon


hellais:
When I click on the ? icon the text overflows and is not readable:

![Screenshot 2019-10-30 at 15 16 24](https://user-images.githubusercontent.com/424620/67865845-4e27b700-fb28-11e9-9066-5221b287d11e.png)

The ideal fix for this would be to follow a pattern similar to how we do it in the mobile app where the card "opens up" and fills the screen.

If we can't do that maybe just putting some scrollbars on the side of it is an acceptable workaround/mvp.

sarathms:
This issue was addressed in #76 and merged to `master` via #74. But leaving it open because this can be fixed better than how it is done now.

hellais:
@sarathms I added this to the cycle backlog. Can you add an effort estimate to this?

sarathms:
This was previously addressed by shoving the long description on the back of those cards into a scrollable div.

![image](https://user-images.githubusercontent.com/700829/70074079-0df4a200-15c8-11ea-8778-3e30469c9888.png)

I have been trying out a couple of approaches, but none seem to land well. Here's one. We make the flipped card full-width and let it occupy as much height as it needs. But not sure if it looks good when the second card is flipped.

<img src="https://user-images.githubusercontent.com/700829/70074289-5d3ad280-15c8-11ea-990d-69d889211b7b.png" width="300px" /> <img src="https://user-images.githubusercontent.com/700829/70074310-72affc80-15c8-11ea-93b7-e28f14a843d2.png" width="300px" />

The other approach I have in mind, which is a bit harder to implement involves some spring animation with `react-spring` which we already use in the project. The idea is to expand the card to full width and height over the other cards and back. But do we even want something like that?

I think we need to think this user interaction through before rushing to implement this. @holantonela I would love to hear your thoughts on this?

sarathms:
Based on my discussion with @hellais, we decided that we will leave these cards the way they are right now. We will address the problem as part of implementing a separate screen where users can perform more actions and see more information similar to how it is done on the mobile apps.

PR #76, opened 2019-11-01
Handle overflowing test card descriptions 

sarathms:
The descriptions are now in scrollable boxes.
Fixes (not really) #63 (Will close it when a proper solution is implemented)


hellais:
Looks good to me. Merging it into the stable-release branch.

PR #77, opened 2019-11-01
No beta

### LINKED ISSUE #72: Replace references to this being a beta
We currently mention in some places in the app that this is a beta.

In preparation for the stable release we should drop all references to it being a beta release.

For example the BETA  button on windows and the fact that the about window opens automatically.

sarathms:
Fixes #72

### DIFF
--- a/main/index.js
+++ b/main/index.js
@@ -64,7 +64,7 @@ if (!isDev && firstRun()) {
 
 let menuTemplate = [
   {
-    label: 'BETA',
+    label: 'About',
     submenu: [
       { label: 'About OONI Probe', click: () => openAboutWindow() },
     ]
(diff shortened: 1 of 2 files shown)


main/index.js
```diff
@@ -146,10 +146,6 @@ app.on('ready', async () => {
       /* eslint-enable no-console */
   }
 
-  if (config._is_beta === true) {
-    // XXX do something like delete all the measurements on a fresh launch.
```
hellais: We may want to keep this and add some extra logic to do some cleanup if we notice the `_is_beta` key inside of the config file.

This is mostly for some possible inconsistencies between the `probe-cli` version of the beta and what we are going to be shipping in the stable release.

@bassosimone may have some thoughts about this.

If this is not necessary we can just drop it directly, but let's also remember to drop that key from the config file.

sarathms: I have restored this piece for now. To be able to land this in the release branch.

PR #80, opened 2019-11-14
Delete old update code, cleanup modern auto-update code

hellais:
There were however left-overs from an older disabled autoupdate system (the default electron one) which needed to be deleted and was causing confusion.

It also wasn’t doing anything with the auto update messages and wasn’t automatically installing the update.

All has been fixed in this PR.


renderer/pages/about.js
```diff
@@ -63,6 +78,7 @@ class About extends React.Component {
           </Box>
         </Flex>
         <Container>
+          {updateMessages.forEach((msg) => (<Text>{msg}</Text>))}
```
sarathms: I'm not sure if this is the best place to render these messages. How about moving this between the `Hard Reset` button and the `Debug information` section?

sarathms: Further:
1. If it is just that one message, then it shouldn't matter.
2. But will this message even be visible, when it is just going to quit and install immediately?

hellais: Yeah I am trying to see if there is a way to test this actually and style it properly.

I think to fully test it we really need to try doing the update process. My theory is that it will be visible because it will also show messages like:
```
Checking for update...
Update available.
Download speed: xxxx
Update downloaded. Quitting and installing.
```

The last message will likely be barely visible, but hopefully the user already figured out this was going on before seeing that last message.

hellais:
I made it look something slightly more reasonable:

Screenshot 2019-11-14 at 16 51 31

PR #82, opened 2019-11-14
OONI Probe Desktop 3.0.0 Release Candidate 3

hellais:
This updates the latest probe-cli version & adds support for showing the ETA in the running test screen.


lang/en.json
```diff
@@ -39,7 +39,7 @@
   "Dashboard.Overview.ChooseWebsites": "Choose websites",
   "Dashboard.Running.Running": "Running:",
   "Dashboard.Running.EstimatedTimeLeft": "Estimated time left:",
-  "Dashboard.Running.Seconds": "{seconds} seconds",
+  "Dashboard.Running.Seconds": "{seconds}",
```
hellais: @sarathms can you update this in the places where it should be updated too?

renderer/components/home/running.js
```diff
@@ -158,10 +156,16 @@ const RunningTest = ({testGroup, logOpen, onToggleLog, progressLine, percent, lo
...
-        <FormattedMessage id='Dashboard.Running.EstimatedTimeLeft' />
-        <FormattedMessage id='Dashboard.Running.Seconds' values={{ seconds: eta }}/>
-      </Text>
+      {eta > 0 &&
+      <Flex justifyContent='center'>
+        <Box pr={1}>
+          <FormattedMessage id='Dashboard.Running.EstimatedTimeLeft' />
+        </Box>
+        <Box>
+          <FormattedMessage id='Dashboard.Running.Seconds' values={{ seconds: moment.duration(eta*1000).humanize() }}/>
```
sarathms: By using `moment.duration.humanize()` we would force strings like 'seconds' and 'minutes' in non-en locales. One solution is to set a locale for the `moment` instance like [here](https://momentjs.com/docs/#/durations/locale/).

lang/en.json
sarathms: We need not translate this string at all. Instead we can drop using that this key. It leads to another problem that I mentioned in the other comment.

PR #83, opened 2019-11-14
Fix width of Raw Data modal

sarathms:
Makes the `Fixed` modal full width. Since the sidebar has a higher `z-index`, we avoid an overlap with it by adding a margin to the left of the modal.

Looks like this now:
image


hellais:
LGTM

Issue #31, opened 2019-06-11
Improvements to the macOS dmg


holantonela:
I cannot find a mockup for the installer, but we can refine this by fixing some minor details:

- [x] the OONI logo is out of the windows. Is intentional on a beta version?
- [x] inner subsection nav makes a double top nav. We should remove both.
- [x] the background looks pixelated. Make sure you are giving support for @2x and @3x retina displays.

<img width="652" alt="Captura de Pantalla 2019-06-06 a la(s) 1 34 08 p  m" src="https://user-images.githubusercontent.com/7602174/59276027-e386e500-8c33-11e9-97f0-22433c159bdd.png">

hellais:
>  the OONI logo is out of the windows. Is intentional on a beta version?
>  inner subsection nav makes a double top nav. We should remove both.

I suspect these two issues are related. Do you perhaps have some special settings to how your windows on macOS look like?

Here is how the DMG looks like when I open it in my computer:

<img width="546" alt="Screenshot 2019-06-12 at 10 46 46" src="https://user-images.githubusercontent.com/424620/59341262-76d91c80-8cff-11e9-99c6-101cc454700a.png">

>  the background looks pixelated. Make sure you are giving support for @2x and @3X retina displays.

You are right. It appears that electron-builder doesn't natively support this, but they link to a stackoverflow issue for a possible workaround. See: https://www.electron.build/configuration/dmg

hellais:
@holantonela I fixed these in: https://github.com/ooni/probe-desktop/pull/84#issuecomment-558760824. Does this looks OK?

hellais:
I gave a little bit more margin to the OONI Probe logo so that even if you have some different Finder settings it should still not cut the logo.

PR #84, opened 2019-11-26
Use tiff to support retina display

hellais:
Set the window size manually to avoid the logo overflowing

This fixes: https://github.com/ooni/probe-desktop/issues/31


hellais:
Here is how it looks like:

Screenshot 2019-11-26 at 19 26 33

PR #88, opened 2019-12-03
Add UI for Psiphon

### LINKED ISSUE #46: Write UI for the Psiphon test result in Desktop
This depends on: https://github.com/ooni/probe-engine/issues/89

sarathms:
This adds UI for Psiphon test. As a dependency, this also has UI for the newly implemented `Circumvention` test group itself.

Fixes ooni/probe#957


hellais:
Some conflicts need to be resolved with this PR before merging.

renderer/components/result/MeasurementRow.js
```diff
@@ -158,18 +139,17 @@ const StatusBox = ({testName, testKeys, isAnomaly}) => {
 // XXX still need to show the summary in here
 const TestRow =  ({measurement, query, testKeys, isAnomaly}) => {
 
-  const icon = iconMap[measurement.test_name]
-  let fullnameID = fullnameMap[measurement.test_name]
-  if (!fullnameID) {
-    fullnameID = 'Test.MISSING.Fullname'
-  }
+  const icon = tests[measurement.test_name].icon && React.cloneElement(
```
hellais: Maybe we want this to not fail if the `test_name` is missing?

sarathms:
Another thing I forgot to flag was that we don't have the animation for `Circumvention` group in here. https://github.com/ooni/design-system/tree/master/components/animations

hellais:
More feedback here:

the icon for the running circumvention test is not the correct one. See:
Screenshot 2020-01-28 at 18 05 02


I think this is because we don't have an animation for it yet, can you use in the meantime as a placeholder the static Circumvention icon?

hellais:
When I click on the row of a result I get the following error:
Screenshot 2020-01-28 at 18 08 16

I suspect this is due to the fact that my ooniprobe is too recent and it's failing on the line I commented on above: https://github.com/ooni/probe-desktop/pull/88#discussion_r371932969

renderer/components/result/MeasurementRow.js
sarathms: This will most likely happen when we try to use `$OONI_HOME` which was created/modified by a `probe-cli` version not supported by `probe-desktop`. If `test_name` is missing, quite a few things will break across the app. If we have to handle all such cases, I suggest we do that in a separate PR.

sarathms: It could be as simple as no loading the measurement for unsupported `test_name`s

sarathms: Opened https://github.com/ooni/probe/issues/986

renderer/components/nettests/circumvention/Psiphon.js
```diff
@@ -0,0 +1,73 @@
...
+import { Text } from 'rebass'
+import { NettestPsiphon } from 'ooni-components/dist/icons'
+
+import colorMap from '../../colorMap'
+
+const Psiphon = ({measurement, isAnomaly, render}) => {
+  const testKeys = JSON.parse(measurement.test_keys)
+
+  // TODO Remove this when actual testKeys come in from newer probe-cli
+  testKeys['bootstrap_time'] = 5.1123455
```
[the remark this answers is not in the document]
sarathms: Removed placeholder. Now rendering from test `measurement.test_keys`

sarathms:
For this PR, hiding results for unsupported tests.

sarathms:
The 'test is running' animation will be replaced with group icons for groups without an animation. But, it looks like the icons don't scale very well.
Screenshot from 2020-01-29 08-32-49

hellais:
I tested this and it's looking good.

We should do a bit more polish to it based on the feedback from @holantonela in here: https://github.com/ooni/probe/issues/967

PR #92, opened 2019-12-16
Hide results for incomplete tests

### LINKED ISSUE #48: Incomplete tests crash Test Results screen
When there are results with `{is_done: false}`, the Test Results listing screen crashes because the incomplete tests appear on top of the older test in the `results` array and breaks the chronology of the tests.

sarathms:
Fixes #48


hellais:
This looks good to me.

We may want to eventually still show the results which are partial so that the user is able to do the cleanup of the results.

For the moment, though I think this is good.

PR #94, opened 2020-01-28
Upgrade styled-components to v4

sarathms:
Upgraded to `styled-components@^4.0.0`
Fixes ooni/probe#898


renderer/components/BarLoader.js
```diff
@@ -17,8 +17,13 @@ const loading = keyframes`
   }
 `
 
+const loadingAnimation = props =>
+  css`
+    ${loading} ${props.duration}s infinite ease-in-out;
+  `
```
hellais: :+1:

PR #95, opened 2020-01-29
Add UI for Tor test

sarathms:
Fixes ooni/probe#975
Implements ooni/probe#967


sarathms:
Screenshot from 2020-01-29 23-33-14

sarathms:
@hellais As per [this](https://github.com/ooni/spec/blob/master/nettests/ts-023-tor.md#expected-output), `tcp_connect` is an array. How do we use this to determine a single :heavy_check_mark: or :x: for the `target` row? cc: @bassosimone 

Is it the same for `tls_handshakes`?

hellais:
@bassosimone knows this better, but I believe that both `tcp_connect` and `tls_handshakes` are lists with a single item inside. The item will have a `status->failure` (for `tcp_connect`) or `failure` (for `tls_handshake`), which if it's NULL, you will show a ✔️ and if it's not NULL, you will show a ❌ with next to it the string.

However there is also another 2 cases, which are those that do the HTTP request and the obfs4 bridges which do a bootstrap.

I think, though we should probably expose these failure strings as top level strings to make it easier for clients to determine the final blocking reason. Maybe we can have two top level keys called `failure_connect` and `failure_handshake`.

@bassosimone what do you think?

bassosimone:
In this specific context they contain just a single entry because we group by target. I am against adding too many top level keys, if possible. My suggestion to process the information in here is to check the first entry in `tcp_connect`, if any, and then the first entry in `tls_handshakes`, if any.

sarathms:
> check the first entry in tcp_connect, if any, and then the first entry in tls_handshakes, if any.

That's how I had [already implemented](https://github.com/ooni/probe-desktop/pull/95/files#diff-f33a6ac09c844fc12a84e2bb4c5a40f8R134) for `tcp_connect`, but wanted to check if there are conditions I'm overlooking.
Anyway, as we discussed over the call, I'll wait for the incoming `summary` values in `test_keys` and populate the last two columns in the table with them.

sarathms:
@hellais Showing the error message along with the red cross in the `Connect` and `Handshakes` column wasn't possible. There isn't enough room for that. Instead I have added a way to toggle the rows to expand and show more content. This also solved the problem of long names in first column.

I need inputs on 2 items:
1. Styling the rows to show that the expanded section is part of the row above it.
2. Is there a better way to indicate that the rows are expandable? Right now, hovering over the entire row turns the cursor to the pointer with a default tooltip saying 'Toggle Expanded' (which I think can be overriden).

image

hellais:
@sarathms I replied to this comment in the issue here: https://github.com/ooni/probe/issues/967#issuecomment-580658797

sarathms:
Now looks like this

sarathms:
Updated copy for a couple of missing strings to cut a release.

PR #98, opened 2020-02-06
UX improvements to test results detais screens

sarathms:
Fixes ooni/probe#990


hellais:
Yes, I think this looks much better.

PR #100, opened 2020-02-07
Explicitly send the SIGINT signal

hellais:
Probe-desktop tentative fix for: https://github.com/ooni/probe/issues/1005.

xref: https://github.com/ooni/probe-cli/pull/106


sarathms:
It doesn't seem to work well on Windows. We are unable to reliably send `SIGINT` to `probe-cli` with the same methods that seem to work on Linux and MacOS.

Maybe we could try using https://www.npmjs.com/package/windows-kill. I can do this later today or tomrorow.

hellais:
This has been superseeded by https://github.com/ooni/probe-desktop/pull/103

PR #102, opened 2020-02-12
Handle incomplete results

sarathms:
Fixes ooni/probe#966


renderer/components/utils.js
```diff
@@ -5,3 +5,13 @@ export const openInBrowser = (url, event) => {
   event.preventDefault()
   shell.openExternal(url)
 }
+
+export const parseTestKeys = (testKeys) => {
+  let parsedTestKeys = null
+  try {
+    parsedTestKeys = JSON.parse(testKeys)
+  } catch (e) {
+    parsedTestKeys = null
```
hellais: Do we maybe want to do something with this error? Perhaps logging it to stdout?

Also is it perhaps not redundant to set this again as null, since you have already initialised it as null?

main/actions.js
```diff
@@ -64,11 +64,13 @@ const listResults = () => {
...
-        // TODO improve how we handle terminating an OONI Probe run in probe-cli
-        if (data.fields.is_done == true) {
-          rows.push(data.fields)
-        }
+        rows.push(data.fields)
+
+        // Sort the result_items because probe-cli returns incomplete results
+        // first and gets pushed at one end of the array
+        rows.sort((resultA, resultB) => (
+          new Date(resultA.start_time) - new Date(resultB.start_time)
```
hellais: Do we need to do something special here to deal with the fact that the time could be in a non UTC locale?

sarathms: As long as both `start_time`s are UTC, the difference would be the same and shouldn't affect the sorting logic.

renderer/components/utils.js
sarathms: I was not familiar with `electron-log`. Didn't want to use `console.log`, so reassigned `null`.

PR #105, opened 2020-02-26
Use custom sorting function for tor connection status columns

sarathms:
Fixes ooni/probe#1007
Allows sorting the icons in columns showing connection status.
image


sarathms:
Something needs to be done for to handle long strings in the `Connect` and `Handshake` columns. It affects the table width.

sarathms:
>  to handle long strings in the `Connect` and `Handshake` columns

This is fixed too.

image

PR #107, opened 2020-02-28
Unsticky Hero on measurement pages

sarathms:
Fixes ooni/probe#959

Only the top header with back button and test name is sticky on the page.

image

@hellais maybe we should have a border or light shadow under the sticky

### DIFF
--- a/renderer/components/measurement/MeasurementContainer.js
+++ b/renderer/components/measurement/MeasurementContainer.js
@@ -12,10 +12,9 @@ import {
 import { FormattedMessage } from 'react-intl'
 import { Tick } from 'ooni-components/dist/icons'
 import { MdPriorityHigh } from 'react-icons/md'
-
+import styled from 'styled-components'
 import { tests } from '../nettests'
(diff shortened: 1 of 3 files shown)


hellais:
I think it works well as-is. Maybe we could add a subtle shadow, but only after you have scrolled beyond the header, which I suspect is non-trivial to do.

The only thing I would fix the z-index of the table elements, see screenshot below:
Screenshot 2020-03-04 at 15 48 37

hellais:
Another thing I noticed in testing is that the header area is not draggable, which is a problem on macOS as there is no other place to drag it from.

It should have the same style as the other draggable areas: https://github.com/ooni/probe-desktop/blob/8601473848fae84a7e993467faa039fb1e48363a/renderer/components/StickyDraggableHeader.js#L14

hellais:
As I am making a release I would like to include at least a basic improvement for this.

I am going to merge this as-is and file a new issue for the follow up fixes: https://github.com/ooni/probe/issues/1033

PR #108, opened 2020-03-14
Enable draggability on header to drag windows on MacOS

sarathms:
Fixes ooni/probe#1033
The zIndex issue was already solved.


hellais:
I tested it on macOS and it works great! Thanks!

hellais:
I merged it (though I clicked the wrong buttons in this UI so it doesn't appear): https://github.com/ooni/probe-desktop/commit/2fce13589656a4489e9acc3f5dede91fb6318c8c

PR #113, opened 2020-03-17
Show OONI logo in measurement screen Hero

sarathms:
Fixes ooni/probe#1042

It looks too small, but larger logo in that position overlaps with the network name.

image

### DIFF
--- a/renderer/components/measurement/MeasurementContainer.js
+++ b/renderer/components/measurement/MeasurementContainer.js
@@ -11,11 +11,12 @@ import {
 } from 'ooni-components'
 import { FormattedMessage } from 'react-intl'
 import { Tick } from 'ooni-components/dist/icons'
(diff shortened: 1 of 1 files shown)


hellais:
As mentioned in https://github.com/ooni/probe/issues/1042#issuecomment-601649566 I think we should not overdo it with branding. Since we have already added the logo to the sidebar, which means it's going to be part of every screenshot anyways, we go for that.

In general I think that we should rather provide "native" share functionality in OONI Probe so that people aren't sharing screenshots, but rather are sharing OONI Explorer links.
That would also assist in ensuring there is adequate attribution and sourcing of the data.

To this end I have filed: https://github.com/ooni/probe/issues/1070

sarathms:
We are not implementing it this way. Closing.

PR #114, opened 2020-03-18
Fix alignment of data usage stats on results screen

sarathms:
Fixes ooni/probe#1045

Had to make some changes to layout to accommodate the larger data usage stats component. Looks like this now:
image

### DIFF
--- a/renderer/components/HumanFilesize.js
+++ b/renderer/components/HumanFilesize.js
@@ -5,7 +5,7 @@ import styled from 'styled-components'
 import humanize from 'humanize'
 
 import {
-  Box
+  Flex
 } from 'ooni-components'
 
 const FileUnit = styled.span`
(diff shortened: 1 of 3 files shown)


hellais:
Can we keep the size of the text itself as it is, but just veritically align it?

I think making it so affects the heirarchy of the elements giving too much importance to the data usage.

That is to say we should still preserve a 3 column layout and make the text size smaller.

As a reference this is how this pattern is done on mobile:

Screenshot_20200320-125142

sarathms:
Even by aligning vertically, the font size (of data usage values) has to be changed to keep the height of the header from expanding too much. In the screenshot you shared, the font size of the data usage values seem half that of the values in first two columns.

sarathms:
image
Does this look reasonable? Haven't committed it yet. It still needs some adjustment to vertically align the first two columns.

hellais:
> Does this look reasonable? 

I think the network and test numbers are still not vertically aligned to the data usage. They should be a bit lower I think.

Can we also try left aligning the data usage icons pointing up and down?

sarathms:
@hellais made some more minor tweaks

Screenshot from 2020-04-07 17-12-57

PR #115, opened 2020-03-18
Add analytics with Matomo

sarathms:
Fixes ooni/probe#1064

image


renderer/pages/settings.js
```diff
@@ -178,36 +178,51 @@ class Settings extends React.Component {
...
+                <NumberOption
+                  onConfigSet={this.reloadConfig}
+                  label={<Text>Websites tested (0 means all)</Text>}
+                  optionKey='nettests.websites_url_limit'
+                  config={config}
+                />
+
+                <BooleanOption
+                  onConfigSet={this.reloadConfig}
+                  label={<Text>Collect analytics data using Matomo</Text>}
```
hellais, line 219: This string is missing from the translations. Can we use a string which we already have in the mobile app which is already translated?

renderer/pages/settings.js
```diff
@@ -178,36 +178,51 @@ class Settings extends React.Component {
...
+
+                <BooleanOption
+                  onConfigSet={this.reloadConfig}
+                  label={<FormattedMessage id='Settings.Sharing.IncludeIP' />}
+                  optionKey='sharing.include_ip'
+                  config={config} />
+
+                <NumberOption
+                  onConfigSet={this.reloadConfig}
+                  label={<Text>Websites tested (0 means all)</Text>}
```
hellais, line 212: This strings is also missing.

renderer/pages/settings.js
```diff
@@ -178,36 +178,51 @@ class Settings extends React.Component {
...
+                  label={<FormattedMessage id='Settings.Sharing.IncludeNetwork' />}
+                  optionKey='sharing.include_asn'
+                  config={config}
+                />
+
+                <BooleanOption
+                  onConfigSet={this.reloadConfig}
+                  label={<FormattedMessage id='Settings.Sharing.IncludeCountryCode' />}
+                  optionKey='sharing.include_country'
+                  config={config}
```
hellais, line 201: While we are at it, let's also remove the ability to disable submitting the country code. See: https://github.com/ooni/probe/issues/1105

sarathms: removed

renderer/pages/settings.js
agrabeli: @hellais We don't have a string for this in the mobile app either. 

Perhaps we can replace the copy of this string with:

Collect app usage statistics

renderer/pages/settings.js
agrabeli: Oh my, this string hasn't been translated? :(

agrabeli: We can replace the copy with:

Number of tested websites (0 means all)

sarathms: added

renderer/pages/settings.js
sarathms: added

PR #116, opened 2020-03-19
Add animation when running Circumvention tests

sarathms:
Fixes ooni/probe#1063


sarathms:
The animation doesn't seem to work in the app yet. When the animation is rendered, the whole react app is unmounted and shows up as a blank white screen. Needs further investigation.

PR #118, opened 2020-03-24
End-to-End testing with Spectron

sarathms:
We use [Spectron](https://github.com/electron-userland/spectron) to run end-to-end tests. Github Actions integration heavily derived from [spectron's own integration](https://github.com/electron-userland/spectron/blob/master/.github/workflows/ci.yml).

Fixes: ooni/probe#1062


sarathms:
Opened ooni/probe#1076 and ooni/probe#1075 to address related pending issues.

PR #119, opened 2020-03-31
Add translations

sarathms:
Fixes ooni/probe#1072

This should enable the app to show the translated strings based on the user's system locale.


hellais:
We documented these follow up issues, but will merge this for the time being:
* https://github.com/ooni/probe/issues/1086
* https://github.com/ooni/probe/issues/1087
* https://github.com/ooni/probe/issues/1088

PR #121, opened 2020-04-01
UI to show that test is stopping

sarathms:
Fixes ooni/probe#1032

Shows an animated progress bar at 99% when the close button is clicked on a running test screen.

image


hellais:
LGTM!

PR #122, opened 2020-04-02
Log to Sentry when autoUpdater thrown an error

sarathms:
Fixes ooni/probe#1068

When `electron-updater` throws an error, we log it to Sentry in the `error` event handler.


hellais:
Documented a follow up issue to this here: https://github.com/ooni/probe/issues/1089

PR #123, opened 2020-04-06
Handle errors in parsing test results

sarathms:
Fixes ooni/probe#1067

### DIFF
--- a/main/actions.js
+++ b/main/actions.js
@@ -3,6 +3,7 @@ const { Ooniprobe } = require('./utils/ooni/ooniprobe')
 
 const log = require('electron-log')
 const debug = require('debug')('ooniprobe-desktop.main.actions')
+const Sentry = require('@sentry/electron')
 
 const hardReset = () => {
   const ooni = new Ooniprobe()
@@ -106,6 +107,11 @@ const showMeasurement = (msmtID) => {
     ooni.on('data', (data) => {
       if (data.level === 'error') {
         debug('error: ', data.message)
+        Sentry.addBreadcrumb({
+          category: 'actions',
+          message: data.message,
+          level: Sentry.Severity.Error
+        })
         reject(data.message)
         return
       }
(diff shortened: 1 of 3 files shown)


renderer/components/test-results/ResultRow.js
```diff
@@ -207,44 +210,58 @@ class ResultRow extends React.Component {
...
-      SummaryElement = summaryMap[name]
+    try {
+      const {
+        name,
+        test_keys,
+        measurement_anomaly_count,
+        measurement_count,
+        is_done
+      } = this.props
+      console.log('renderTestKeys...')
```
sarathms, line 245: remove this

renderer/components/test-results/ResultRow.js
```diff
@@ -172,30 +158,33 @@ class ResultRow extends React.Component {
...
+              </Box>
+              <Box>
+                <Text color={group.color} bold>{group.name}</Text>
+              </Box>
+            </Flex>
+          </Box>
+        </Flex>
+      )
+    } catch (e) {
+      Sentry.captureException(new Error('name missing in result', e))
```
hellais, line 185: Will this capture also the exception backtrace?

Maybe instead of creating a new Error, which may loose the exception stacktrace, we should log the error which you have caught in the catch and add a breadcrum with the selected testGroup name?

https://docs.sentry.io/enriching-error-data/breadcrumbs/?platform=electron#recording-crumbs

main/actions.js
```diff
@@ -53,15 +53,17 @@ const listMeasurements = (resultID) => {
...
   let rows = [],
+    errors = [],
     summary = {}
 
   return new Promise((resolve, reject) => {
     ooni.on('data', (data) => {
-      if (data.level === 'error') {
+      if (data.fields.type === 'result_item'  && data.level === 'error') {
         log.error('listResults: error in row', data.message)
-        reject(data.message)
```
hellais, line 62: Should we maybe log these to sentry as well?

hellais:
LGTM

PR #124, opened 2020-04-07
Use ndt7

sarathms:
Fixes ooni/probe#1040

Didn't have to change anything in `NDT.js`. Just had to fallback to using test_keys from `ooniprobe list <id>` output instead of using the one from the raw measurement. This was earlier done to render the table in the Tor msmt details. Now reverted and uses another approach.

### DIFF
--- a/renderer/components/useRawData.js
+++ b/renderer/components/useRawData.js
@@ -1,0 +1,24 @@
+import React from 'react'
+import { useRouter } from 'next/router'
+import electron from 'electron'
+const debug = require('debug')('ooniprobe-desktop.renderer.components.hooks.useRawData')
+import Raven from 'raven-js'
+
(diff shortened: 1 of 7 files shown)


renderer/components/useRawData.js
```diff
@@ -0,0 +1,24 @@
...
+import Raven from 'raven-js'
+
+export const useRawData = () => {
+  const [rawData, setRawData] = React.useState(null)
+  const [error, setError] = React.useState(null)
+  const { query } = useRouter()
+  const remote = electron.remote
+  const { showMeasurement } = remote.require('./actions')
+
+  React.useEffect(() => {
```
hellais, line 14: Interesting. I didn't know of this hook. It's cool!

PR #130, opened 2020-04-14
Tests in jest

sarathms:
Fixes ooni/probe#1078

* Switched to writing async tests in Jest
* Added custom sequencer to test onboarding process first and also resetting home directory before onboarding.
* Added directory to collect screenshots (maybe there is a way to organize the screenshots)

_Note: The tests themselves are not exhaustive._


hellais:
🐳

PR #133, opened 2020-04-16
Fix result summary spacing

sarathms:
Fixes ooni/probe#1043
Fixes ooni/probe#1110


hellais:
The vertical alignment of the elements is still a bit off:
Screenshot 2020-04-20 at 12 39 55

I am merging this, but we should not close ooni/probe#1043 as it's not fully resolved yet.

PR #134, opened 2020-04-17
UI polish

sarathms:
Fixes ooni/probe#1098


hellais:
Merging, but documenting the things which are not fixed in a dedicated issue.

hellais:
https://github.com/ooni/probe/issues/1113

PR #137, opened 2020-04-20
Add icon next to ExternalLink

hellais:
Fixes one of the points in ooni/probe#1098


renderer/components/ExternalLink.js
```diff
@@ -1,12 +1,16 @@
...
+
 import { Link } from 'ooni-components'
 
 import { openInBrowser } from './utils'
 
 const ExternalLink = ({href, color = 'blue5', children, ...props}) => (
   <Link {...props} color={color} href={href} onClick={(e) => openInBrowser(href  , e)}>
-    {children}
+      {children}
+      <MdOpenInNew />
```
sarathms, line 13: Indentation

```suggestion
    {children}
    <MdOpenInNew />
```

sarathms:
I've added the change to pass attributes like `title` as prop to `<ExternalLink>`

PR #145, opened 2020-05-04
WIP: Add Continuous Delivery

Ceda-EI:
- [x] Do Linux Builds
- [ ] Do Windows Builds
- [ ] Do MacOS Builds

Example builds were run on the test-ci branch of my fork. 

- [Pre Release](https://github.com/Ceda-EI/probe-desktop/releases/tag/v3.0.1-rc.1)
- [Release](https://github.com/Ceda-EI/probe-desktop/releases/tag/v3.0.1)


Ceda-EI:
The failing CI seems to be unrelated.

hellais:
This is really cool! Thanks for putting it together!

You are right that the CI build on macOS has been causing quite some issues, as a matter of fact I am now going to disable it until we find why it's not working well (see: https://github.com/ooni/probe/issues/1157).

I don't know so much about linux app distribution, so maybe @FedericoCeratto or @sarathms can look at that part and see if it makes sense to integrate this.

WRT the macOS and windows builds, the challenge with doing CD on those is that we would have to give our code signing keys to github, which I think we should first carefully discuss and evaluate the tradeoffs of that approach.

That said, maybe we can just have unsigned developer builds made using this process and keep the "standard" process for signed builds.

This PR relevant to: https://github.com/ooni/probe/issues/1091.

Ceda-EI:
I actually raised the concern about the need of signing keys to be uploaded to github in the IRC. On top of that, Github does not have any mechanism for uploading secret files into the build. Github's recommended approach is committing an encrypted version of the key to repo and then adding the password to it as a secret. [Source].

[Source]: https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#limits-for-secrets

package.json
```diff
@@ -29,6 +29,7 @@
     "appId": "org.ooni.probe-desktop",
     "productName": "OONI Probe",
     "asar": true,
+    "artifactName": "${productName}.${ext}",
```
sarathms, line 32: It is useful to have the version number in filename.

PR #146, opened 2020-05-05
Avoid unnecessary about window popup

sarathms:
Skips popping up about window to notify 'update (not) avaialble'

### DIFF
--- a/main/index.js
+++ b/main/index.js
@@ -119,16 +119,11 @@ function sendUpdaterProgress(progressObj) {
   aboutWindow.webContents.send('update-progress', progressObj)
 }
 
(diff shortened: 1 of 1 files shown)


main/index.js
```diff
@@ -119,16 +119,11 @@ function sendUpdaterProgress(progressObj) {
   aboutWindow.webContents.send('update-progress', progressObj)
 }
 
-autoUpdater.on('checking-for-update', () => {
-  sendStatusToWindow('Checking for update...')
-})
```
hellais, line 124: Maybe we should have logged it to the console?

PR #147, opened 2020-05-08
Show measurements in local timezone

sarathms:
Fixes ooni/probe#1099

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "description": "OONI Probe Desktop app",
   "author": "Open Observatory of Network Interference (OONI) <contact@openobservatory.org>",
   "productName": "OONI Probe",
-  "version": "3.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "description": "OONI Probe Desktop app",
   "author": "Open Observatory of Network Interference (OONI) <contact@openobservatory.org>",
   "productName": "OONI Probe",
-  "version": "3.0.1",
+  "version": "3.0.2",
```
hellais, line 6: We should not bump the version number until we are ready to make a release.

Perhaps a good practice is to, after we make a release, replace the version number with `${probeVersion}-dev`

hellais:
I think we should disable the macOS build until we figure out why it's failing.

I made an issue for it: https://github.com/ooni/probe/issues/1157

PR #155, opened 2020-06-24
Fix/translation related

sarathms:
Fixes ooni/probe#1204

Also updates psiphon details UI a bit.

becomes

because of [this change](https://github.com/ooni/probe-desktop/pull/155/files#diff-173d4409de80068aad12540a182437b9L260-R274)


hellais:
hmm, what was the reason for putting the second unit inside of the to be translated string? Is it so that some languages are able to put the unit before or after the number?

It does look a bit better with the unit being styled differently, but it's still OK.

Let's merge this and maybe take note of this caveat in the translation related improvements issue: https://github.com/ooni/probe/issues/1140.

PR #156, opened 2020-06-26
Show confirmation dialog before stopping test

sarathms:
Fixes ooni/probe#1205
Screenshot from 2020-06-26 01-19-14

I used the same style we use in the onboarding screens. I tried using the buttons from our design system, but dropped it later.
Screenshot from 2020-06-26 00-47-24

Also fixes the bug where the currently running test name wasn't being shown.
Screenshot from 2020-06-26 01-19-08

### DIFF
--- a/renderer/components/Modal.js
+++ b/renderer/components/Modal.js
@@ -1,0 +1,38 @@
+import React from 'react'
+import { Fixed, Box } from 'ooni-components'
+import styled from 'styled-components'
+
+export const Modal = styled(Fixed)`
+  z-index: 1001;
+  max-width: 100vw;
+  max-height: 100vh;
+  overflow: auto;
+  transform: translate(-50%, -50%);
+  box-shadow: rgba(0, 0, 0, 0.80) 0 0 0 60vmax, rgba(0, 0, 0, 0.25) 0 0 32px;
+  border-radius: 10px;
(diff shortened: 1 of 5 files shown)


hellais:
We should be using a different modal styling for this.

The pattern for the modal in the onboarding is specific for the quiz.

Here is what needs to change:

* Invert the buttons to have OK on the right and cancel on the left.
* Make the background color of the modal be white
* Primary button (OK) should be blue background with white text, secondary button (cancel) should be blue on white
* Text should all be black

The copy also is a bit weird. Did @agrabeli review it?

hellais:
Here is a quick mockup of how it should look like:

Screenshot 2020-06-29 at 17 51 44

Important things to keep in mind for the implementation:
- Adding shadow around the modal box to give sense of depth
- Ensure that the tap target around the cancel button is as bug as the OK button and it highlights when mouse over (i.e. it's actually a button with a white background and blue text, like the other one)

sarathms:
Are we not using the `Button` style from `ooni-components`?

hellais:
> Are we not using the Button style from ooni-components?

Yes, if you already have a button component, use that.

hellais (approved):
This looks great!

For me it's good to go!

🐳

PR #164, opened 2020-07-24
Improved Settings

sarathms:
Fixes ooni/probe#1088
Fixes ooni/probe#1022
Fixes ooni/probe#1121

- [x] Refactor settings components as functional components
- [x] Update layout to new mockups in ooni/probe#1125
- [x] Add a LangaugeSelector MVP
- [x] Show supported languages in selector dropdown


hellais:
I see that when I change the language there are some strings that have still boilerplate in them, why is that?
Screenshot 2020-09-01 at 10 17 32

hellais:
The category code selection needs a bit more work:

Screenshot 2020-09-01 at 10 18 33

- The top right corner close icon should receive the same treatment as the modal in the stop test UI
- The subtle fade effect doesn't work well and I think it's affecting the buttons as well, let's remove it and document it as future work

- At the end of the page there is unnecessary whitespace:
Screenshot 2020-09-01 at 10 20 02

- When I close the modal after having made some changes, I would expect them to not be saved, while in reality they are saved

hellais (changes requested):
Needs a bit more polishing. Left some feedback in comments.

sarathms:
image

sarathms:
@hellais I refactored the modal code to share the code between this screen and the 'stop-the-test' screen. Please test that screen also a few times along with this.

@agrabeli Can you review/edit [the copy](https://docs.google.com/spreadsheets/d/1GPkr_OyNhXJRTXnseyNbo2P4Du-i_W9kJdqjZG4HVkM/edit#gid=0) for this confirmation popup? Keys are like `Settings.Websites.Categories.*`. If/when you edit it, do notify me here or on slack, so that I can include those changes to this PR. Thanks!

sarathms:
@hellais Please note that I changed how we write the categories into the config file. Instead of an array, it is a dictionary/map. Reasons : *) will be easier for someone trying to edit the file manually *) made it easier to write some code
```
{
  "_version": 3,
  "_informed_consent": true,
  "sharing": {
    ...
  },
  "nettests": {
    "websites_url_limit": 10,
    "websites_enabled_category_codes": {
      "ALDR": true,
      "ANON": true,
      "COMM": false,
      "COMT": true,
      "CTRL": false,
```
On a related note, in this change we are bumping the config file version because we added the categories entry.

main/utils/config.js
```diff
@@ -5,28 +5,62 @@ const log = require('electron-log')
 
 const { getHomeDir } = require('./paths')
 
-const LATEST_CONFIG_VERSION = 2
+const LATEST_CONFIG_VERSION = 3
 
 const OONI_CONFIG_PATH = path.join(getHomeDir(), 'config.json')
 
+const availableCategoriesList = {
```
hellais, line 12: This should be a list

main/utils/config.js
```diff
@@ -5,28 +5,62 @@ const log = require('electron-log')
...
+    'sharing': {
+      'include_ip': false,
+      'include_asn': true,
+      'upload_results': true
     },
-    "nettests": {
-      "websites_url_limit": 0
+    'nettests': {
+      'websites_url_limit': 0,
+      'websites_enabled_category_codes': availableCategoriesList
```
hellais, line 56: You can do something like: 
```
Object.keys(availableCategoriesList).reduce(k => availableCategoriesList[k] === true)
```

hellais:
This is starting to look good!

Screenshot 2020-11-09 at 11 58 20

A few small things that I think should be addressed:
- The scroll bar is scrollable in the background of the modal (see above screenshot). It should instead not be scrollable when the modal is displayed.
- Unsaved changes are displayed even if no change was made
- The unsaved changes modal should replace the current modal and not be displayed as an overlay
- The fonts in the modal look a bit weird. Could it be that some CSS effect is applied to them?

hellais:
I did some light testing of it and functionally it's working as expected.

sarathms:
> The unsaved changes modal should replace the current modal and not be displayed as an overlay

@hellais Isn't it a common pattern to show this kind of save-or-discard confirmation prompt before taking down the UI where the changes were made? It seems counter-intuitive despite the fact it is a modal over a modal.

hellais:
> Isn't it a common pattern to show this kind of save-or-discard confirmation prompt before taking down the UI where the changes were made? 

It's generally considered an anti-pattern to show a modal within another modal, see:
* https://ux.stackexchange.com/questions/52042/is-it-acceptable-to-open-a-modal-popup-on-top-of-another-modal-popup
* https://ux.stackexchange.com/questions/120182/display-modal-over-another-modal-its-really-a-sin

If we really want to keep the modal we should either:
* Remove the one below and just show the confirm modal
* Keep both, but make the second focussed by fading out the one in the background by putting it in the same faded out background

I don’t think it’s really needed to have a confirm are you sure to cancel modal, since it’s not really such an important thing to save. Perhaps we can make it clearer that the done button needs to be clicked on in order to save the data, by renaming it from DONE to SAVE.

That way it’s clear that the X is to cancel, while the DONE button is needed in order to commit the state to the disk.

sarathms:
@hellais What exactly is the problem with fonts? I am yet to notice anything. I checked for CSS effects or overrides, nothing applies. All the text that I checked has the right fonts applied.

hellais:
It's hard to describe. They look a bit "fuzzy" like there is some odd aliasing effect going on, but maybe it's an optical illusion or a brainfart on my end.

sarathms:
Category Code Selections
image

Reordered sections
image

Localized names of languages
image

hellais (approved):
LGTM

PR #167, opened 2020-08-24
Use probe-cli-3.0.7-beta.1

sarathms:


bassosimone (approved):
🐳

sarathms:
Merging the beta version so that we can test it with other PRs to catch possible bugs. We will update to a stable version tag before releasing it.

PR #168, opened 2020-08-27
Fix UI flicker when running tests

sarathms:
Fixes ooni/probe#1189

@hellais Can you check how this works on MacOS?


bassosimone:
@sarathms could you please fix the conflicts and rebase this PR on master so that I can test it? (I'm also taking the liberty of marking this as ready for review since a review is basically the thing we need to do now :-)

sarathms:
@bassosimone Rebased. Merge was done right, as per UI also.

bassosimone:
I run

```
git pull
git checkout fix/flicker
yarn
yarn probe-cli
yarn dev
```

and got the following error when running web connectivity:

```
Error: Objects are not valid as a React child (found: Error: failed with code 2). If you meant to render a collection of children, use an array instead.
    in p (created by Running)
    in div (created by Context.Consumer)
    in StyledComponent (created by running__StyledRunningTest)
    in running__StyledRunningTest (created by Running)
    in div (created by Context.Consumer)
    in StyledComponent (created by running__ContentContainer)
    in running__ContentContainer (created by Running)
    in div (created by Context.Consumer)
    in StyledComponent (created by running__WindowContainer)
    in running__WindowContainer (created by Running)
    in Running (created by Home)
    in div (created by ForwardRef)
    in ForwardRef (created by Context.Consumer)
    in StyledComponent (created by Root)
    in Root (created by Rebass.Provider)
```

bassosimone:
@sarathms I've just merged master into this branch and I am taking another look.

renderer/pages/home.js
```diff
@@ -1,5 +1,5 @@
 /* global require */
-import React from 'react'
+import React, { useState, useCallback, useEffect } from 'react'
```
bassosimone, line 2:
```suggestion
import React, { useState, useCallback } from 'react'
```

seems unused in this file?

renderer/components/home/running.js
```diff
@@ -236,63 +147,210 @@ const ContentContainer = styled.div`
...
+        debug('error', error)
+        Raven.captureException(error, {
+          extra: { scope: 'renderer.runTest' }
+        })
+        setError(error)
+      })
+
+
+    return () => {
+      const { ipcRenderer } = require('electron')
```
bassosimone, line 200:
```suggestion
```

Why do you import `ipcRenderer` again in this inner scope? Isn't `ipcRenderer` already part of the closure of this function because you already import it above?

renderer/pages/home.js
bassosimone: (Is there a tool we can run to detect and fix these issues automatically?)

bassosimone (approved):
Thank you for organising the PR in commits each with a different scope. It has been very useful to help me understand what you were doing.

🐳 with a couple of small suggestions.

renderer/components/home/running.js
sarathms: removed

renderer/pages/home.js
sarathms: removed. We already have `eslint` which should detect this. My atom setup seems to be messed up and hasn't been showing this in some cases.

PR #169, opened 2020-08-27
Replace abandoned react-lottie package with react-lottie-player

sarathms:
Fixes ooni/probe#1218 

Used for the animations during the onboarding quiz and test running screen.


bassosimone (approved):
🐳

PR #174, opened 2020-09-08
Fix spacing and alignment of result overview contents

sarathms:
Fixes ooni/probe#1043
spacing between icons and labels.
image

Also went ahead and increased the vertical spacing between the rows. The screen looks like this. 

image


bassosimone (approved):
🐳

PR #176, opened 2020-09-17
fix(scripts/download-bin.js): don't download the zip archive

bassosimone:
Spotted while doing a review for https://github.com/ooni/probe/issues/1218


sarathms (approved):
:whale: You are right. This is better.

PR #177, opened 2020-09-30
package.json: use ooni/probe-cli@v3.0.7

bassosimone:
Part of https://github.com/ooni/probe-engine/issues/871


sarathms (approved):
Build seems to run as intended. We can bundle this. :whale:

PR #178, opened 2020-10-05
Fix bug in checking animation data validity

sarathms:
Fixes ooni/probe#1253

There was some left over code from #169


bassosimone (approved):
🐳

PR #179, opened 2020-10-05
package.json: use ooni/probe-cli@v3.0.8

sarathms:
Please note that we should merge #178 first before any releases.


bassosimone (approved):
It's WAI on macOS, windows, and linux, so I'd say we can merge! 🐳

PR #181, opened 2020-10-27
update: ooni/probe-engine v3.0.9

bassosimone:
See https://github.com/ooni/probe-engine/issues/972


sarathms (approved):
:whale: Runs fine with `probe-cli@3.0.9`

PR #182, opened 2020-11-11
New Dashboard

sarathms:
Fixes ooni/probe#1274

- [X] New layout for test cards
- [x] New details screen for each test
- [x] Choose websites (~blocked by ooni/probe#1283~)
  - [x] Editor to add custom list of websites
  - [X] Run `websites` test with custom list via `--input-file`
- [X] `Run All` button
- [x] Run all tests function
- [x] Page based navigation through detail pages
- [x] Custom callback hook to launch a test from anywhere


sarathms:
@hellais Can you try out the work done here so far? Particularly the 'Run All Tests' feature. It is functional now and works pretty much the same way it works on mobile.
image


When the `all` test is launched, there is a unpredictable amount of delay before we can start showing the running test animation screen (name, progress, ETA etc). We cannot directly start showing any _one_ of the test groups because the order in which the tests run seems to be random. Also we do not know which test has started running until we get the first progress message. 

To fill this gap, for now, I have mapped the group `all` to `default`, an old [fallback group](https://github.com/ooni/probe-desktop/blob/master/renderer/components/nettests/index.js#L35-L41). We can replace the name and icon to resemble a loading screen and show them before the first animation.
image

Alternatively, we can load up a custom lottie animation similar to https://lottiefiles.com/16240-octopus until the first test group animation can be shown.

hellais:
@sarathms do we have a spinner we can use for this first screen?

I would replace the copy to say:

"Starting all tests"

and drop the "Running" sentence from below

I will put some time to make a mock for this in the next sprint. I say we aim to do this for the next release 2.2.0.

sarathms:
@hellais I rebased this branch with the latest master after the 3.1.0 release. I didn't realize you had already pulled this once. Please nuke your local branch and pull again when you checkout next time.

hellais:
Some comments:
* We should tighten up the padding between the cards on the landing page so that we can get a peak of the middlebox test card
* For the moment we can disable the choose websites button so we can make this ready to ship
* The run button should not have the greyed out effect on hover
* The go back button from the test card detail view should be the right cursor type

sarathms:
Rebased this branch onto latest master for the new probe-cli@v3.1.0 that provides `--input-file` feature that we plan to use to implement the 'Choose Websites` feature on desktop.

sarathms:
As of now, we have a working 'choose websites' feature, without the test list editor. Using IPC, the renderer process requests the main process to create a file with the test list content. The main process replies with the path to the file it created in `${TEMP}`. We then go to `/dashboard/running?testgroupname=websites&inputFile=/path/to/temp/file` which launches the test.

sarathms:
We now have the problem that by enabling max runtime the way it was done, the entire "run all" session gets terminated.

sarathms:
Discussed with @bassosimone and we concluded that it is probably optimal to make `probe-desktop` call `ooniprobe run ${group}` on each test group instead of running `ooniprobe run all` to avoid this problem.

hellais (approved):
I did some light testing of the branch and it looks good to me! 🐳

sarathms:
I added a fix to a problem that happens only in production builds. Going ahead with merge.

PR #183, opened 2020-11-13
chore: use probe-cli 3.0.10

bassosimone:
See https://github.com/ooni/probe-engine/issues/1004


sarathms (approved):
Works fine on the desktop. Thanks @bassosimone :whale:

PR #185, opened 2020-11-20
Feature: Delete All Measurements

sarathms:
Fixes ooni/probe#1275


sarathms:
Progress so far

image

image

Yet to align with the mockups in ooni/probe#1267

- [ ] button styling
- [ ] Calculating size of the correct directory `${OONI_HOME}/msmts/`
- [ ] Confirmation dialog styling

hellais:
Let's just just use the same style as the EDIT button above for the "Delete all measurements" button

sarathms:
1. ~What is the styling decision behind the `Delete All` button? Is it a link or a differently styled `<Button hollow>`~
This was already answered above.
2. When the user performs the action, I think the expectation is to see a zero value size as a result. This is possible only if we calculate the size of `${OONI_HOME}/msmts`. This seems like the logical thing to do. Let me know if there is something else also to consider.

hellais:
Let's align the copy to the mobile app.

Instead of "Total storage used" it should read:
"Storage used" and the button below should read "Clear"

hellais:
> When the user performs the action, I think the expectation is to see a zero value size as a result. This is possible only if we calculate the size of ${OONI_HOME}/msmts. This seems like the logical thing to do. Let me know if there is something else also to consider.

If we can do this and obtain a consistent result, the config file is regenerated properly and the DB will be re-created as expected, then go for it.

However there is the issue that we are doing it only for uploaded measurements, because I believe we don't yet have support for checking which are the uploaded measurements and re-uploadeding the unuploaded.

I think we can do this incrementally and in the first iteration not care about the total not going to zero, but accept that, if there are some un-uploaded measurements for it to be non-zero.

PR #186, opened 2020-11-21
Check for null value in category code selection during config migration

sarathms:
Fixes ooni/probe#1290


sarathms:
@bassosimone has reviewed this code during call with me and we agreed to merge this PR in the current state to enable pre-release testing. He will be reviewing again and testing this part on Mac. If we find something, we fix it in a separate PR.

bassosimone:
:shipit:

PR #190, opened 2020-12-15
Limit test duration

sarathms:
Fixes ooni/probe#1280

* Adds a migration function that
  * removes the url_limit option
  * adds two new keys `websites_enable_max_runtime` and `websites_max_runtime`
* Use setTimeout to start a timer in node.js to kill the `child_process`
* Cancel the timer when manually stopped in UI
* Disable the numeric input when the first boolean option is turned off

image

Disabled 
image


main/utils/ooni/run.js
```diff
@@ -42,6 +63,9 @@ class Runner {
         value: logMessage
       })
     })
+
+    this.maybeStartMaxRuntimeTimer()
```
hellais, line 67: Since this is an async call, don't you need to await on it before returning?

hellais:
When max_runtime is enabled the Estimated time left should be that of the maxruntime and not that returned by probe-engine:
Screenshot 2021-01-06 at 14 13 47

sarathms:
@hellais I have adressed the progress bar and ETA comments. I noticed a bug where the config context wasn't updated after onboarding. It seemed to happen only in that context. But the fix I added basically checks if the value in `ConfigContext` is `null` and fetches the config to refresh the context. This check happens whenever `useConfig` hook is called.

hellais:
The progress bar and ETA are not working as expected.

The progress keeps flickering between the runtime given out by probe-engine and the overridden progress.

Moreover the status update messages are not being displayed anymore.

See at around 1m of the video:

https://user-images.githubusercontent.com/424620/104020438-8495a300-51bd-11eb-8a40-cb31f14fec19.mov

hellais (approved):
I did another pass and this looks good to me! 🐳

PR #191, opened 2021-01-12
chore(package.json): use ooni/probe-cli 3.2.0

bassosimone:
See https://github.com/ooni/probe-engine/issues/1125


sarathms (approved):
Works fine with desktop :whale:

PR #195, opened 2021-02-01
Optimize build size

sarathms:
This attempts to reduce the production build size further from what was done along with the `3.1.0` release. 

~It also upgrades the core dependencies like `electron` to latest versions.~ I had to upgrade electron to `11.x` and then downgrade back to `8.x` because there were breaking changes in core areas like IPC that need to be addressed before we can safely upgrade.

Otherwise this improves Sentry integration, upgrades nextjs to the latest version and gets rid of a lot of unused dependencies.

This also addresses the next.js deprecation notice about `static` dir as described [here](https://github.com/vercel/next.js/blob/master/errors/static-dir-deprecated.md).

Fixes ooni/probe#1151


sarathms:
Currently, sentry configuration is broken.

renderer/components/IntlProvider.js
```diff
@@ -19,14 +19,14 @@ const IntlProvider = ({ children }) => {
...
-  require(`@formatjs/intl-displaynames/locale-data/${activeLang}`)
+  // require(`@formatjs/intl-displaynames/locale-data/${activeLang}`)
 
   const messages = getMessages(activeLang)
 
   const intl = createIntl({ locale: activeLang, messages }, cache)
 
   const changeLocale = (locale) => {
-    require(`@formatjs/intl-displaynames/locale-data/${locale}`)
+    // require(`@formatjs/intl-displaynames/locale-data/${locale}`)
```
hellais, line 29: These should not be commented out.

hellais (approved):
LGTM

renderer/pages/_app.js
```diff
@@ -7,7 +7,8 @@ const MyApp = ({ Component, pageProps, err }) => {
 
   // Workaround for https://github.com/zeit/next.js/issues/8592
   const modifiedPageProps = { ...pageProps, err }
-  initSentry()
+
+  // initSentry()
```
hellais, line 11: Why is this commented out?

renderer/components/initSentry.js
```diff
@@ -1,32 +1,48 @@
 import * as Sentry from '@sentry/node'
 import { RewriteFrames } from '@sentry/integrations'
+// import { ipcRenderer } from 'electron'
```
hellais, line 3: This should be uncommented and reverted the `require` import call.

hellais (approved):
LGTM

PR #196, opened 2021-02-03
chore: use ooni/probe-cli v3.5.1

bassosimone:
Part of https://github.com/ooni/probe-engine/issues/1181


sarathms (approved):
:whale: thanks!

PR #197, opened 2021-02-11
Change default values at end of onboarding process

sarathms:
At end of onboarding when 'Change defaults' is clicked, we set the config to opt-out of analytics. Problem with letting the user perform the change action is that a config file is already written by then and it would have 'opted-in' to allow the user to opt-out and then stands the risk of staying opted-in if the app is closed or navigated away. We should discuss if this is how we want to do this.

### DIFF
--- a/renderer/pages/settings.js
+++ b/renderer/pages/settings.js
@@ -1,5 +1,5 @@
-import React, { useEffect, useState, useCallback } from 'react'
-import electron from 'electron'
+import PropTypes from 'prop-types'
+import React, { useMemo } from 'react'
 import { FormattedMessage } from 'react-intl'
 import styled from 'styled-components'
 import {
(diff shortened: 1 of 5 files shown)


renderer/pages/settings.js
```diff
@@ -78,6 +86,10 @@ const Settings = () => {
                 label={<FormattedMessage id='Settings.Privacy.CollectAnalytics' />}
                 optionKey='advanced.collect_usage_stats'
               />
+              <BooleanOption
+                label={<FormattedMessage id='Settings.Privacy.SendCrashReports' />}
+                optionKey='advanced.send_crash_reports'
+              />
               <BooleanOption
                 label={<FormattedMessage id='Settings.Sharing.UploadResults' />}
                 optionKey='sharing.upload_results'
```
hellais, line 95: While we are at it, we should also remove the option to enable IP address collection.

hellais:
> Problem with letting the user perform the change action is that a config file is already written by then and it would have 'opted-in' to allow the user to opt-out and then stands the risk of staying opted-in if the app is closed or navigated away

Won't they in this case still be presented with the onboarding when the re-open the app? Can we change the point in which the config file is written so that it only happens when their either click on "Let's Go" or "Change defaults"?

sarathms:
> Won't they in this case still be presented with the onboarding when the re-open the app? Can we change the point in which the config file is written so that it only happens when their either click on "Let's Go" or "Change defaults"?

We do write the file only after the user makes a choice between those two options. I was pointing to the fact that in both cases we record consent i.e `informed_consent: true`

hellais (approved):
LGTM

PR #198, opened 2021-02-12
Automatic testing on MacOS and Windows

sarathms:
Fixes ooni/probe#1350
Fixes ooni/probe#1356


sarathms:
Rebasing with latest master.

main/index.js
```diff
@@ -31,14 +37,11 @@ require('electron-debug')({
...
 autoUpdater.logger = log
 autoUpdater.logger.transports.file.level = 'info'
-log.info('App starting...')
+
 
 // To prevent garbage collection of the windows
 let windows = null
 
-// Set the application name
-app.name = 'OONI Probe'
```
hellais, line 40: Why was this removed?

main/utils/autorun/schedule.js
```diff
@@ -0,0 +1,57 @@
+const { join } = require('path')
+const { initConfigFile } = require('../config')
+const { getAutorunHomeDir } = require('../paths')
+
+// Initialize OONI_HOME for autorun by generating a fresh config file
+// in the OONI_HOME for autorun
+// Note: initConfigFile in turn initializes Sentry (again) with no extra effect
+initConfigFile({ configFilePath: join(getAutorunHomeDir(), 'config.json') })
```
hellais, line 8: How is the config file for the autorun kept in sync with the config file for the normal OONI Probe?

main/utils/autorun/mac-scheduler.js
```diff
@@ -0,0 +1,55 @@
...
+        OONI_HOME_autorun: getAutorunHomeDir()
+      })
+      const taskPlistPath = join(
+        app.getPath('home'), 'Library', 'LaunchAgents',
+        `${taskname}.plist`
+      )
+      writeFile(taskPlistPath, taskPlistStr).then(() => {
+        log.debug(`Autorun task plist file created: ${taskPlistPath}`)
+        log.debug(`Enable task with: launchctl load ${taskPlistPath}`)
+        const cmd = launchctl('load', [taskPlistPath])
```
hellais, line 36: Are you sure this is the right command? In probe-cli we are instead doing:
```
launchctl enable && launchctl bootstrap
```

main/utils/autorun/mac-scheduler.js
```diff
@@ -0,0 +1,55 @@
...
+        log.debug(`Enable command result: ${cmd.stdout.toString().trim()}`)
+        resolve()
+      }).catch(e => {
+        reject(`Failed to create plist file: ${e.message}`)
+      })
+    })
+  },
+  delete: function (taskname) {
+    return new Promise((resolve, reject) => {
+      // launchctl('unload', [taskPlistPath])
```
hellais, line 51: Here it should delete the file and call `bootout`

main/utils/autorun/taskTemplateWin.js
```diff
@@ -0,0 +1,73 @@
...
+
+const getStartBoundary = () => {
+  const now = new Date()
+  // In one minute
+  now.setMinutes(now.getMinutes() + 1)
+
+  // use Date.toISOString() and account for timezone offset
+  // StartBoundary format - YYYY-MM-DDTHH:mm:ss
+  // Date.toISOString()   - YYYY-MM-DDTHH:mm:ss.sssZ
+  now.setMinutes(now.getMinutes() - now.getTimezoneOffset())
```
hellais, line 11: Why are you calling `setMinutes` twice?

main/index.js
sarathms: We will use the value `productName` in `package.json` to keep it consistent. It already uses the same value.

main/utils/autorun/schedule.js
sarathms: I lost sight of all the usecases for which we wanted to `sync` and not `copy`. I have to do this now.

main/utils/autorun/mac-scheduler.js
sarathms: In most places I read, it is recommended to use `load` because it registers the task and starts it right away.
https://medium.com/swlh/how-to-use-launchd-to-run-services-in-macos-b972ed1e352
https://www.maketecheasier.com/use-launchd-run-scripts-on-schedule-macos/
https://apple.stackexchange.com/questions/29056/launchctl-difference-between-load-and-start-unload-and-stop

main/utils/autorun/mac-scheduler.js
sarathms: This is now done.

main/utils/autorun/mac-scheduler.js
hellais: @bassosimone maybe we want to use the same behaviour on probe-cli then.

bassosimone: I think we should not. The `load` command is deprecated and may not work. I think it's safer that `probe-desktop` does what `probe-cli` does, which is what `brew services` does.

sarathms: Thanks @bassosimone for showing how `load/unload` is not the recommended way to do this. I will go ahead and change this.

hellais (approved):
LGTM

PR #199, opened 2021-03-04
chore: use probe-engine 3.6.0

bassosimone:
See https://github.com/ooni/probe/issues/1340


sarathms:
We will be using v3.7.0 instead in #201. Closing this one.

PR #202, opened 2021-03-15
UI for Signal Test

sarathms:
Fixes ooni/probe#1408
Fixes ooni/probe#1358


sarathms:
https://github.com/ooni/probe-desktop/blob/master/renderer/components/nettests/circumvention/NettestRiseupVPN.svg
https://github.com/ooni/probe-desktop/blob/ux/signal/renderer/components/nettests/im/NettestSignal.svg

sarathms:
@bassosimone This branch should ideally have `probeVersion: 3.6.0` but I already merged `v3.7.0` into `master` because that is what we will be shipping soon. I am testing it by locally using `3.6.0`. I will either merge it as-is to `master` and later bump `probeVersion` when you have a new probe-cli with Signal test in `im` group, or I can keep this open until such a release and then merge to `master`. Not a big deal, but still checking if you have thoughts about it.

hellais:
I think I might have to export it using Illustrator to work in the same way. Can you try with this SVG and see if it works better:

[NettestSignal.svg.zip](https://github.com/ooni/probe-desktop/files/6141709/NettestSignal.svg.zip)

sarathms:
Looks like this will work.

hellais:
I tested this branch using a probe-cli build from: https://github.com/ooni/probe-cli/pull/259 and it's looking good.

We just need the strings to be added to the translation files, get them translated and then we should be good to go.

sarathms:
Rebased with master. At this point, we cannot run the signal test or show it because the latest `probe-cli@3.9.0` that allows to do this, won't work because it needs swapping of `stdout` and `stderr` which will be done in another PR.

Merging this now.

sarathms:
image

image

image

PR #203, opened 2021-03-16
Restore window on MacOS when activated from dock

sarathms:
Currently, when the app window is closed, it doesn't quit and shows active in the dock. When clicked, it doesn't show the app. This fix completes this behavior by creating and showing the window.

Additionally, it also implements the single instance restriction as recommended by [electron docs](https://www.electronjs.org/docs/api/app?q=unhandled#apprequestsingleinstancelock).


hellais (approved):
LGTM

PR #206, opened 2021-03-31
RTL Support

sarathms:
Closes ooni/probe#1087

- [x] CSS configuration
- [x] Include RTL language translations
- [x] Identify locales to enable RTL layouts for
- [x] Address community feedback by fixing styles for reported components


sarathms:
Feedback from community:

- [ ] Content in about page need not be in RTL
- [ ] In general, content that is not translated (e.g coming from probe-cli) need not be in RTL
- [ ] Logs are missing altogether (in Windows)
- [ ] Progress bar when running tests seems to be flipped even for non-RTL languages

renderer/components/IntlProvider.js
```diff
@@ -13,6 +13,8 @@ require('@formatjs/intl-displaynames/locale-data/en')
 // https://formatjs.io/docs/react-intl/components#rawintlprovider
 const cache = createIntlCache()
 
+const localesWithRTL = ['ar', 'fa']
```
hellais, line 16: We might as well add all the possible RTL language codes in here: https://github.com/shadiabuhilal/rtl-detect/blob/master/lib/rtl-detect.js#L138

hellais (approved):
LGTM

PR #208, opened 2021-04-05
Remove unused options from config.json

sarathms:
* Removes the matomo-based usage analytics from the app.
* Remove all unused values from `config.json` 
* Fixes ooni/probe#1373 
  But the code is part of initializing analytics which this PR removes from the app.


hellais (approved):
LGTM

PR #221, opened 2021-06-01
feat: new onboarding screen for crash reporting optin

sarathms:
Fixes ooni/probe#1455

Screen Shot 2021-06-01 at 3 30 11 PM


hellais (approved):
LGTM

PR #225, opened 2021-06-10
Component tests for Sections.js, QuizSteps.js and Stepper.js

Eddges:
I'm currently trying to write tests that simulate how a user would interact with that component. I'm not sure if this is the correct/desired way. Kindly review @sarathms 

* I haven't written tests for `QuizSteps.js` and Screen 3 (Crash Reporting) of `Sections.js` component yet because of some hurdles. Getting to the solution of either of the two will solve it for both. 
* For `Stepper.js`, I figured writing snapshot tests makes the most sense

* I think writing some functions in a different file and importing them will be a good idea, for example `rgbToHex()` or `renderComponent()`. I thought of creating a `utils` directory to store such functions, but felt I should ask you first. Kindly provide suggestion.

### DIFF
--- a/renderer/components/onboard/__tests__/QuizSteps.test.js
+++ b/renderer/components/onboard/__tests__/QuizSteps.test.js
@@ -1,32 +1,31 @@
 /**
  * @jest-environment jsdom
  */
-
+import React from 'react'
 import {
   screen,
   render,
   fireEvent,
-  cleanup,
   waitForElementToBeRemoved,
-} from "@testing-library/react";
-import { theme } from "ooni-components";
-import { ThemeProvider } from "styled-components";
-import { IntlProvider, FormattedMessage } from "react-intl";
-import English from "../../../../lang/en.json";
+} from '@testing-library/react'
+import { theme } from 'ooni-components'
+import { ThemeProvider } from 'styled-components'
+import { IntlProvider, FormattedMessage } from 'react-intl'
+import English from '../../../../lang/en.json'
 
-import QuizSteps from "../QuizSteps";
+import QuizSteps from '../QuizSteps'
 
-const renderComponent = (component, locale = "en", messages = English) => {
+const renderComponent = (component, locale = 'en', messages = English) => {
   return render(
     <IntlProvider locale={locale} messages={messages}>
       <ThemeProvider theme={theme}>{component}</ThemeProvider>
     </IntlProvider>
-  );
-};
+  )
+}
 
-describe("Tests for QuizSteps", () => {
-  const toggleQuiz = jest.fn();
-  const onQuizComplete = jest.fn();
+describe('Tests for QuizSteps', () => {
+  const toggleQuiz = jest.fn()
+  const onQuizComplete = jest.fn()
 
   beforeEach(() => {
     renderComponent(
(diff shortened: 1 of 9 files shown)


Eddges:
I wrote some tests for `QuizSteps.js`, but I'm not sure if it's done in the best way.
The challenge I encountered with this component in particular is due to the nature of animations on clicking 'True' for 'False' in the Pop Quiz. Upon clicking either of those buttons, an animation is triggered that lasts for a second before getting to the next screen. So to access the next screen, I had to resolve to using timers, which makes the tests run longer. Mocking the animation seemed like a better idea but upon finishing of these animations trigger callbacks that change internal states of the component. I wasn't able to figure out a way to mock the component.
Kindly provide feedback.

sarathms:
I need some more time to wrap my head around this, but meanwhile some things to consider:

> but the end of these animations trigger callbacks that change internal states of the component. I wasn't able to figure out a way to mock it.

When I run `QuizSteps.test.js`, I get this warning/error that seems to indicate a way to deal with components which undergo state updates.
```
    Warning: An update to null inside a test was not wrapped in act(...).

    When testing, code that causes React state updates should be wrapped into act(...):

    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */

    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://fb.me/react-wrap-tests-with-act
```
The link leads to documentation for react test-utils. I am not sure whether it is still a recommended tool for testing and if it is OK to be used along with `@testing-library/react`, but the latter seems to have an [`act` API](https://testing-library.com/docs/react-testing-library/api#act) of its own. I came across this which seems something related to look into. https://github.com/skidding/react-mock/tree/master/packages/state

> So to access the next screen, I had to resolve to using timers, which makes the tests run longer.

I have seen some tests using the [watiFor](https://testing-library.com/docs/dom-testing-library/api-async/#waitfor) API to achieve this. Check if you can use it too.

> For Stepper.js, I figured writing snapshot tests makes the most sense

Good call!

> I think writing some functions in a different file and importing them will be a good idea, for example rgbToHex() or renderComponent(). I thought of creating a utils directory to store such functions, but felt I should ask you first. Kindly provide suggestion.

I'm OK with such refactoring in general. For smaller utility functions like `rgbToHex` I encourage you to go ahead and make the changes in the PR and seek review. I am OK if you find some of the components also need some restructuring which we could discuss before diving into the code.

sarathms:
At some point, you can consider moving the definition for `renderComponent` into a separate file which can be imported into all tests? Let me know if I am missing some minor differences in the way it is defined in the 3 files so far.

Eddges:
Thank you for your pointer. I had tried working with `waitFor` but it timed out everytime. I later got to know that manually setting timeout interval was possible and hence I was able to improve the code.

renderer/components/onboard/__tests__/QuizSteps.test.js
```diff
@@ -0,0 +1,108 @@
...
+  })
+
+  afterEach(() => {
+    onQuizComplete.mockClear()
+  })
+
+  test('User Story with correct answers', async () => {
+    const yesButton1 = screen.getByText(English['Onboarding.PopQuiz.True'])
+    fireEvent.click(yesButton1)
+    const lottiePlayerFirst = screen.getByTestId('quiz-steps-animation')
```
sarathms, line 66: This doesn't differentiate between `tickAnimation` and `crossAnimation`. Maybe the `data-testid` can be assigned dynamically inside the `<Animation>` component and it can be tested here (and the in the next test case)

renderer/components/onboard/__tests__/QuizSteps.test.js
```diff
@@ -0,0 +1,108 @@
...
+      English['Onboarding.PopQuiz.2.Wrong.Paragraph']
+    )
+    const continueButton2 = screen.getByText(
+      English['Onboarding.PopQuiz.Wrong.Button.Continue']
+    )
+    expect(warningParagraph2).toBeInTheDocument()
+    fireEvent.click(continueButton2)
+    expect(onQuizComplete).toHaveBeenCalledTimes(1)
+  })
+})
```
sarathms, line 108: Wondering if it is useful to consider writing tests for all 4 combinations of button clicks. e.g [True, False] and [False, True]. What do you think? @Eddges

sarathms (approved):
I think this PR is mostly good to go. I have left a couple of comments for you to respond to.

renderer/components/onboard/__tests__/QuizSteps.test.js
Eddges: Yes, you are correct. Thank you for your suggestion! The test now covers both tick and cross animations.

renderer/components/onboard/__tests__/QuizSteps.test.js
Eddges: I do not see any issue in that. It would rather enhance confidence in the tests. The tests are written for all four combinations now.

sarathms (approved):
👍

PR #230, opened 2021-06-26
[Fix] Input URL fields need not be dirty for validation

Eddges:
There are a couple of changes proposed in this PR: 
1. `validateURL` function has been shifted up so that it is declared first before a reference to it is made. I was getting error in `jest` that said a reference to the function was made before it was declared.
2. The condition that URL fields need to be dirty before calling the validation function has been removed. Now any change that happens to the textbox is validated.

The reason for this change is because of unexpected behavior arising in presence of this condition.
* The `Run` button would not enable unless focus was shifted to some other element in a headful environment. In a headless environment, the button wouldn't enable at all.
* Validation would not trigger unless focus was shifted to some other element.

The behavior is reproduced here: 
https://user-images.githubusercontent.com/32863230/123514663-b342ff80-d6b1-11eb-994c-e90eea8e6225.mp4

This behavior is fixed by removing the condition

Screenshot from 2021-06-26 18 50 30
Screenshot from 2021-06-26 18 50 54

### DIFF
--- a/renderer/components/dashboard/URL.js
+++ b/renderer/components/dashboard/URL.js
@@ -45,36 +45,6 @@ const RemoveButton = styled(Button)`
 
 const URL = ({ idx, url, error, onUpdate, onRemove, onKeyEnter }) => {
   const [dirty, setDirty] = useState(false)
-  const onChange = useCallback((e) => {
-    onUpdate(idx, e.target.value)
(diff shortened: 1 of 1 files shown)


sarathms (approved):
This seems fine to me,

PR #233, opened 2021-07-19
Week 2 tests

Eddges:
I have modified the `Sidebar.js` component to make use of `useRouter()` from `next/router`. There were some bugs I faced on simulating click interaction. On reading about the issues, I came to know that `next/link` component that we use currently has some bugs and is not a great API, having inconsistencies in its implementation.

References: 
https://github.com/vercel/next.js/issues/16864#issuecomment-700593327
https://github.com/vercel/next.js/issues/1490#issuecomment-289108262


sarathms:
I am yet to checkout this branch and run the app. But one thing I would like to point out is that not using `next/link` and treating them as regular `<a>` links may have some impact on how the app handles page transitions. It may or may not be applicable to the way we use `<Link>` in our app. E.g In some cases, clicking `<Link>` only causes client-side changes without a full page reload starting with blank white page.

Eddges:
I understand. I've actually replaced `Link` with `Flex` from `ooni-components` which I think is a regular `div` tag. I have passed an `onClick` listener to it.

renderer/components/__tests__/Sidebar.test.js
```diff
@@ -0,0 +1,81 @@
...
+    expect(screen.getByText('Test paragraph')).toBeInTheDocument()
+  })
+
+  test('NavItem renders Navigation List correctly in the Sidebar', async () => {
+    Object.keys(navigationPaths).map((path, idx) => {
+      const info = navigationPaths[path]
+      renderComponent(
+        <NavItem
+          key={idx}
+          pathName="/dashboard"
```
sarathms, line 64: Why is `/dashboard` hard coded here?

renderer/components/__tests__/Sidebar.test.js
```diff
@@ -0,0 +1,81 @@
...
+  test('Displays child elements', async () => {
+    renderComponent(
+      <Sidebar>
+        <p>Test paragraph</p>
+      </Sidebar>
+    )
+    expect(screen.getByText('Test paragraph')).toBeInTheDocument()
+  })
+
+  test('NavItem renders Navigation List correctly in the Sidebar', async () => {
```
sarathms, line 58: Can we not directly test by rendering `<Sidebar>` and checking if it has all the items in navigationPath and if each of them navigates to the right page?

renderer/components/dashboard/__tests__/DashboardHeader.test.js
```diff
@@ -0,0 +1,32 @@
...
+
+describe('Tests for DashboardHeader', () => {
+  const onRunTest = jest.fn()
+
+  test('Run button is rendered and triggers the test', async () => {
+    renderComponent(<DashboardHeader onRunAll={onRunTest('all')} />)
+    const runButton = screen.getByText(English['Dashboard.Overview.Run'])
+    expect(runButton).toBeInTheDocument()
+    fireEvent.click(runButton)
+    expect(onRunTest).toHaveBeenCalledTimes(1)
```
sarathms, line 30: If we are only testing if the callback was called, then we should be testing the hook `useRunTest` also.

renderer/components/dashboard/__tests__/RunTestCard.test.js
```diff
@@ -0,0 +1,80 @@
...
+
+describe('Tests for RunTestCard component', () => {
+  const router = useRouter()
+
+  afterEach(() => {
+    cleanup()
+    jest.clearAllMocks()
+  })
+
+  test('All the Test Cards are mounted', async () => {
```
sarathms, line 44: This test is relevant only if we are testing by rendering `pages/dashboard/index.js` and checking if all there are cards for each item in `testList`. Otherwise we are manually mounting all the cards in the test setup itself.

renderer/components/dashboard/__tests__/RunTestCard.test.js
```diff
@@ -0,0 +1,80 @@
...
+      <RunTestCard
+        onClick={() =>
+          router.push('/dashboard/[testGroup]', `/dashboard/${websiteDetails.key}`)
+        }
+        id={websiteDetails.key}
+        {...websiteDetails}
+      />
+    )
+
+    const testCard = screen.getByTestId('card')
```
sarathms, line 75: Also test if `testCard` has visible content that matches what's in `websiteDetails`

sarathms (changes requested):
I have left a few suggestions as inline comments to improve some of the tests.

I would encourage to avoid writing tests where the only assertion is `expect(text).toBeInTheDocument()` especially where `text` is not a result of a specific DOM query. For example the Sidebar test for version number. Feel free to sprinkle as many `data-testid`s as you want and use them to query in such simple tests.

Eddges:
I have removed some tests which I felt were not relevant at all. I have only kept the ones which I felt added confidence in the component.

sarathms (approved):
👍 I think this is good to go!

PR #234, opened 2021-07-20
E2E tests covering Onboarding process, Dashboard, Settings and Test Results pages

Eddges:


sarathms (commented):
Premature review. Left some comments anyway.

test/dashboard.e2e.js
```diff
@@ -1,26 +1,108 @@
...
+        desc: 'Test your network speed and performance',
+      },
+      {
+        id: 'middlebox',
+        name: 'Middleboxes',
+        desc: 'Detect middleboxes in your network',
+      },
+    ]
+
+    testDetails.forEach((itr) => {
```
sarathms, line 52: Is this the same as using the [`test.each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout) API from Jest?

test/runtest-all.e2e.js
```diff
@@ -0,0 +1,311 @@
...
+  })
+
+  afterAll(async () => {
+    await stopApp(app)
+  })
+
+  test('Tests start successfully on click of RUN button', async () => {
+    await app.client.$('button[data-testid=button-dashboard-run]').click()
+
+    await waitFor(
```
sarathms, line 20: Why is this pattern being used in all the new `.e2e.js` files? It seems like a practice from `testing-library`. Maybe you can explain to me separately how this is supposed to work. I am only wondering if there is a way to avoid writing so much of repetitive code all around.

test/runtest-individual.e2e.js
```diff
@@ -0,0 +1,612 @@
+import { startApp, stopApp, screenshotApp } from './utils'
+import { waitFor } from '@testing-library/dom'
+
+jest.setTimeout(600000)
```
sarathms, line 4: Is there a reason behind this number?

test/runtest-individual.e2e.js
```diff
@@ -0,0 +1,612 @@
...
+      async () => {
+        const animationVisible = await app.client.isVisible(
+          'div[data-testid=running-animation-websites]'
+        )
+        return expect(animationVisible).toBe(false)
+      },
+      { timeout: 300000 }
+    )
+
+    await app.client.pause(2500)
```
sarathms, line 190: Why is there a pause here?

test/settings.e2e.js
```diff
@@ -0,0 +1,240 @@
...
+      .$('select[data-testid=select-language]')
+      .getValue()
+    expect(languageSelectValue).toBe('en')
+
+    // Clicking and selecting value instead of using using setValue() because
+    // webdriver thinks the <select> option is in a wrong state (uneditable)
+    await app.client.$('select[data-testid=select-language]').click()
+    await app.client.$('option=Spanish').click()
+
+    const languageLabelEs = await app.client.isVisible('label=Idioma')
```
sarathms, line 51: This can be `isESLabelVisible` or something. Then reading the assertion in the next line makes sense.

sarathms: Or use a different assertion than `toBe(true)`.

renderer/components/measurement/ExplorerURLButton.js
```diff
@@ -10,7 +10,7 @@ const ExplorerURLButton = ({ reportID, input }) => {
     explorerURL = `${explorerURL}?input=${input}`
   }
   return (
-    <Button onClick={(e) => openInBrowser(explorerURL, e)}>
+    <Button onClick={(e) => openInBrowser(explorerURL, e)} data-testid='button-show-in-explorer' href={explorerURL}>
```
sarathms, line 13: Why do we need the `href` attribute here?

sarathms (changes requested):
Thanks for expanding the E2E tests to this extent. I can feel the confidence our Quality Assurance grow significantly with this.

I have left a few questions to help me understand some parts better.

I think one key aspect missing from these tests is the Autorun feature. I understand this is because you are developing this test on Linux where we don't have that feature. The only reason I haven't approved it is that, on MacOS (and maybe on Windows also), the `test/test-results.e2e.js` test fails because the Autorun reminder pops up while the test runner is expecting to see something else.
image

My recommendation is to write a `test/autorun.e2e.js` which tests the feature and handles that reminder popup and run it before the `test-results.e2e.js` test.

sarathms:
@Eddges I also found that using this can automatically generate a screenshot at the end of every single test case.
```javascript
  beforeEach(() => {
    screenshotApp(app, expect.getState().currentTestName)
  })
```

Maybe you can add this in all the files in the top-level `describe` block and remove all the screenshotApp calls placed at the end of test-cases. But you can keep the ones that capture screenshots during the test.

package.json
```diff
@@ -25,9 +25,10 @@
     "publish:win": "yarn run build:clean && electron-builder --win -p always",
     "icon-gen": "icon-gen -i static/Icon-Color.svg -o main/static/icons -r -m ico,icns",
     "lint": "eslint --ignore-path .gitignore .",
-    "test:current": "jest TestGroupInDetail.test.js -c test/lib/unit.config.js",
+    "test:current": "jest settings.e2e.js --testRegex 'e2e.js' -c test/lib/integration.config.js",
```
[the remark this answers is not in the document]
Eddges: I understand. I will be removing `yarn run test:current` before making a push from here on to avoid these conflicts. The command doesn't belong in the merged code anyway!

package.json
```diff
@@ -25,9 +25,9 @@
     "publish:win": "yarn run build:clean && electron-builder --win -p always",
     "icon-gen": "icon-gen -i static/Icon-Color.svg -o main/static/icons -r -m ico,icns",
     "lint": "eslint --ignore-path .gitignore .",
-    "test:current": "jest TestGroupInDetail.test.js -c test/lib/unit.config.js",
     "test:unit": "jest -c test/lib/unit.config.js",
     "test:e2e": "jest -c test/lib/integration.config.js --testRegex 'e2e.js' --runInBand",
+    "test:unit:update_snapshot": "jest -c test/lib/unit.config.js --updateSnapshot",
```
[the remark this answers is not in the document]
Eddges: Got it! Thanks!

test/runtest-all.e2e.js
Eddges: I actually find the `waitFor` api to be the easiest one to implement to assert if a promise is resolved or rejected with a given timeout value.

test/runtest-individual.e2e.js
Eddges: I just wanted to set the timeout to a large value in case the CI run takes time.

test/runtest-individual.e2e.js
Eddges: This is because in case of a router event, when a new page is loaded, `webdriverIO` keeps looking for matching DOM nodes in the previous page itself. Pausing for a certain amount of time and then performing DOM matching makes `webdriver` look up in the new page.

test/settings.e2e.js
Eddges: Changed to `isEsLabelVisible`

renderer/components/measurement/ExplorerURLButton.js
Eddges: I needed to find a way to get access to the generated link to browse in explorer. I'm not sure if this a good way though.

test/dashboard.e2e.js
Eddges: Yes, it is

PR #235, opened 2021-07-21
Some tests for language translations

Eddges:


sarathms (approved):
This assumes that the `data/en.csv` file is update with the ids in the code. Although this isn't entirely foolproof, it is consistent with the manual methods in which we update those files. So we can consider this sufficient to merge for now.

A good improvement would be to use the babel plugin for `react-intl` to extract all the ids used in the code (which happens as part of `yarn build`) and use those to test if the translated files cover all those ids. We have [this configuration in place for explorer](https://github.com/ooni/explorer/blob/master/.babelrc#L5-L9)  (but used differently). @Eddges Maybe you can create an issue to implement this for probe-desktop and label it as `post-gsoc` for you to pick up later.

sarathms:
I added a commit to remove the `test:current` script from package.json to avoid an unnecessary merge conflict.

PR #243, opened 2021-08-13
E2E tests - Git history fix

Eddges:
This PR is in response to the messed up Git history of #234


sarathms (approved):
Thank you @Eddges for quickly arranging for a clean branch. These tests are good to go in. We already discussed over the call about future improvements to be documented as issues.

PR #249, opened 2021-10-08
Add support for experimental test group

sarathms:
Fixes ooni/probe#1389

- [x] Card on dashboard
- [x] Icon
- [x] Animation
- [x] Results list view
- [x] Measurement list view
- [x] Generic measurement view with just raw measurement

image
image
image
image

image

image


hellais:
Once the additional file is removed it looks good to merge.

PR #251, opened 2021-10-12
Add language support for Swahili and Dutch

sarathms:
Fixes ooni/probe#1773

 - Add support for Swahili and Dutch
 - Update all translations

image

image

image


sarathms:
I have tested all the screens for both languages and the app displays everything with the translated strings.

PR #255, opened 2021-10-13
Fix init config bug and E2E tests

sarathms:
This fixes a bug in how config file is initialized which in turn was one of the reasons for E2E tests to fail.

Another fix for tests to run correctly is to disable opening the about window before main window when the app launches.


sarathms:
Merging this to ensure we are able to run these tests against the code in `master`

PR #260, opened 2021-11-24
chore: use ooniprobe-cli 3.11.0-beta.2

bassosimone:
This diff changes ooniprobe-desktop to use cli v3.11.0-beta.2.

This cli release includes changes in the last (at this point six!)
months to improve measurements quality.

It is missing some extra fixes required to bless a stable
release. Chiefly among them, the possiblity of running
cleanly the DNSCheck experiment.

I'm going to work on these issues as soon as possible.

In the meanwhile, this release is a good testing base to check
whether we have additional lingering issues.

A few notes to explain this diff follows.

First, I took the liberty of reducing as much as possible
the amount of markdown linting warning I did see.

Second, 3.11.0-beta.2 is the first release that is fully built
using the cloud. I have not added any PGP key to the cloud
as this seems a bit futile. Maybe it is not and maybe I'm not
seeing the full picture, so we should discuss this topic. (I
would not put my PGP key in there, and instead I'd put another
key that has no password to sign releases; is this better
than not using any key at all?)

Third, cloud builds do not use tar.gz anymore. I did this to
work around https://github.com/ooni/probe/issues/1884 (in general,
less pieces we use, less errors could occur.)

As a result, the download script needed a bit of reworking.

Fourth, while I reckon this is unlikely, I couldn't help noticing
that the previous script did not download in case files were
already on disk. To add a bit of robustness to this process, I
am now prefixing the downloaded file with its version. This way,
we'll re-download if there is a cli version change and there
are already files on disk and, by some other mistake, the build
rules are such that we don't clean existing files.

The related issue is https://github.com/ooni/probe/issues/1879

### DIFF
--- a/scripts/download-bin.js
+++ b/scripts/download-bin.js
@@ -13,28 +13,24 @@ const dstDir = path.join(appRoot, 'build', 'probe-cli')
 
 const download = () => {
   ensureDirSync(dstDir)
-  const osarchlist = ['darwin_amd64', 'linux_amd64', 'windows_amd64']
-  for (let i = 0; i < osarchlist.length; i += 1) {
-    const osarch = osarchlist[i]
-    const tarball = `ooniprobe_${osarch}.tar.gz`
-    const tarballURL = `${baseURL}/${tarball}`
-    const sig = `${tarball}.asc`
-    const sigURL = `${tarballURL}.asc`
-    if (existsSync(`${dstDir}/${tarball}`) === false) {
-      console.log(`Downloading ${tarball}`)
-      execSync(`curl -#f -L -o ${dstDir}/${tarball} ${tarballURL}`)
+  const osarchs = [
+    "darwin-amd64",
+    "linux-amd64",
+    "windows-amd64",
+  ]
+  for (let i = 0; i < osarchs.length; i += 1) {
+    const extension = (osarchs[i] === "windows-amd64") ? ".exe" : ""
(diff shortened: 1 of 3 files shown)


scripts/download-bin.js
```diff
@@ -13,28 +13,24 @@ const dstDir = path.join(appRoot, 'build', 'probe-cli')
...
-      execSync(`gpg --quiet --verify ${dstDir}/${sig} ${dstDir}/${tarball}`)
-    } catch (e) {
-      console.error(`Signature verification failure: ${e}`)
+  const osarchs = [
+    "darwin-amd64",
+    "linux-amd64",
+    "windows-amd64",
+  ]
+  for (let i = 0; i < osarchs.length; i += 1) {
+    const extension = (osarchs[i] === "windows-amd64") ? ".exe" : ""
```
bassosimone, line 22: Maybe this could be better written as "if it contains windows then add extension". How do I do "if it contains" in JS?

sarathms: I added a commit that does this. (+ a few other JS stuff and lint fixes).

sarathms (commented):
Thanks @bassosimone! The changes look good to me. I added a commit to do some JS stuff you asked for in a comment.

I am yet to test a desktop app build with this version of `probe-cli`. I will come back here and update the review when I have done that tomorrow.

sarathms (approved):
I tested the build with this version of probe-cli and it works as expected 🐳

PR #267, opened 2021-12-15
feat: use ooni/probe-cli 3.13.0 and expose dnscheck

bassosimone:
This diff uses ooni/probe-cli 3.13.0. I tested that this
specific version was working in linux/amd64.

While there, I noticed that dnscheck was not exposed,
for understandable reasons (it has been broken until the
release of 3.13.0) and exposed it. (It has been quite a
pleasing experience to be able to expose it by just
adding a single line of code for declaring it, well done!)

Reference issue https://github.com/ooni/probe/issues/1843

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "author": "Open Observatory of Network Interference (OONI) <contact@openobservatory.org>",
   "productName": "OONI Probe",
   "version": "3.6.1",
-  "probeVersion": "3.13.0-beta",
+  "probeVersion": "3.13.0",
   "main": "main/index.js",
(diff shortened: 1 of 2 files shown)


bassosimone:
Since writing this diff, I also tested whether everything was working in macOS and it looked fine.

sarathms (approved):
Thank you @bassosimone. Looks good to merge.

Exposing the `dnscheck` test part makes this complete. Although I will be going ahead and removing those parts to address [this issue](https://github.com/ooni/probe/issues/1822) to make it easier to add/remove tests in the experimental group.

PR #271, opened 2022-01-24
feat: add support for torsf

bassosimone:
Reference issue: https://github.com/ooni/probe/issues/1917

### DIFF
--- a/renderer/components/nettests/index.js
+++ b/renderer/components/nettests/index.js
@@ -77,7 +77,8 @@ export const tests = {
   dnscheck: minimalTest('dnscheck'),
   'default': {
     'name': 'Default',
-  }
+  },
+  torsf: minimalTest('torsf')
 }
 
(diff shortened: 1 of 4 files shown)


sarathms (approved):
👍 Looks good to me.

renderer/components/nettests/index.js
```diff
@@ -77,7 +77,8 @@ export const tests = {
   dnscheck: minimalTest('dnscheck'),
   'default': {
     'name': 'Default',
-  }
+  },
+  torsf: minimalTest('torsf')
```
hellais, line 81: nitpick: maybe this item should go before the `default` catch all test_name?

sarathms: The order itself doesn't affect lookups. Although from a readability angle, it makes sense to logically leave the `default` entry at the bottom. I ignored this because I plan to remove all test-specific code for `experimental` anyway.

PR #272, opened 2022-01-28
Fix RiseupVPN result screen issue

sarathms:
Fixes ooni/probe#1953

When the output of `ooniprobe show <id>` fails to return a valid measurement in response, the pages that consume parts of the raw measurement were crashing. This PR fixes that.

First noticed for certain `riseupvpn` tests recently. Fix applied to other places with similar usage.


hellais (approved):
LGTM

PR #273, opened 2022-02-09
Fix crash when removing websites in URL editor

sarathms:
Fixes ooni/probe#1848
﻿
Bug introduced because a valid string like `""` was not being used to
initialize a newly added URL input and the length of such a string was
being used in a condition.


hellais (approved):
LGTM

PR #274, opened 2022-02-10
Fix bug in falling back to en locale correctly

sarathms:
Fixes ooni/probe#1914

### DIFF
--- a/renderer/components/langUtils.js
+++ b/renderer/components/langUtils.js
@@ -1,8 +1,6 @@
-/* global require */
 import osLocale from 'os-locale'
 
 const defaultLocale = 'en'
-const defaultLocaleName = 'English'
 
 export const getMessages = (locale = null) => {
   let supportedMessages = {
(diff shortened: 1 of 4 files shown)


renderer/components/langUtils.js
```diff
@@ -13,15 +11,15 @@ export const getMessages = (locale = null) => {
...
 
-  if (supportedMessages.hasOwnProperty(locale)) {
+  if (locale in supportedMessages) {
     const mergedMessages = Object.assign(
       {},
       supportedMessages[defaultLocale],
       supportedMessages[locale]
     )
     return mergedMessages
   } else {
```
hellais, line 21: Maybe you can even drop the `else` statement here, to make it more readable and robust.

hellais (commented):
Small nitpick, once it's addressed it can be merged.

hellais (approved):
LGTM!

PR #275, opened 2022-02-11
chore: use probe-cli 3.14.0-beta

bassosimone:
Where applicable, also re-enable torsf.

See https://github.com/ooni/probe/issues/1845.


sarathms (approved):
This is good to go.

I have made some changes to the end-to-end tests 
- disabled tests related to riseupvpn
- added tests for the experimental card. But unlike other groups, we won't be testing for specific tests within (e.g TorSF, dnscheck) to allow flexibility. We do test that the raw data is available for these tests.

PR #276, opened 2022-02-11
Fix conflict with probe-cli in using max_runtime

sarathms:
probe-desktop uses `nettests.websites_enable_max_runtime` from the config and ignores `nettests.websites_max_runtime` when running websites card.

Whereas probe-cli always uses `nettests.websites_max_runtime` (unless it is `0`) to limit the runtime.

This means even if it is disabled in the desktop UI, runtimes are still limited by `probe-cli` because it sees a non-zero value in `config.json`.

The currently proposed solution forces `nettests.websites_max_runtime` to `0` when it is disabled. This loses our ability to remember if the user had previously set a custom duration before disabling.

One way to overcome that would be to store that value in `settings.json` and retrieve it. But I think that adds unnecessaryredundancy.

The ideal solution is probably to move all `probe-desktop` settings to `settings.json` and use it to generate a `config.<timestamp>.json` whenever probe-cli is launched.

### DIFF
--- a/renderer/components/settings/widgets.js
+++ b/renderer/components/settings/widgets.js
@@ -28,15 +28,18 @@ const StyledErrorMessage = styled(Box).attrs({
   color: ${props => props.theme.colors.red5};
 `
 
-export const BooleanOption = ({ label, optionKey, disabled = false, ...rest }) => {
+export const BooleanOption = ({ label, optionKey, disabled = false, onChange, ...rest }) => {
   const [checked, setConfigValue] = useConfig(optionKey)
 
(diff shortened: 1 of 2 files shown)


sarathms:
Probably, this also fixes ooni/probe#1833

sarathms:
For now, we will set the max_runtime to a default (e.g `90`) when the checkbox is enabled. The proper solution i.e using `settings.json` to store these values and generating a on-the-fly `config.json` for `probe-cli` will be implemented separately.

sarathms:
We additionally hide the numeric option when the boolean option is unchecked.

PR #277, opened 2022-02-14
Upload E2E test screenshots + Update tor download links

sarathms:
- Upgrade unit test deps
- Upload screenshots from E2E tests as workflow artifacts
- Use latest valid link to download tor binaries

I ended up updating the structure of one of the unit test files while doing this. This needs to be replicated across all unit test files to ensure that the test environment setup is uniform across the testing process.


sarathms:
@hellais I am going ahead and merging this to land the latest download link for tor binaries in `master`. Without that, CI fails for all other PRs and prevents tests from running. I have opened an issue (ooni/probe#2028) to address this properly.

PR #285, opened 2022-03-02
Do not initialize Sentry before initializing config

sarathms:
There was a circular dependency requiring the modules `main/utils/sentry.js` and `main/utils/config.js` from `main/index.js`. This fix breaks it by lazily loading one of the modules when needed.


hellais (approved):
LGTM

PR #290, opened 2022-03-07
Update about window launch configuration

sarathms:
Uses the same `BrowserWindow` configuration for About window and main window.


sarathms:
This was already present in the branch `release/3.7.0`

PR #295, opened 2022-05-07
Delete ooni_home_autorun on hard reset

majakomel:
Closes https://github.com/ooni/probe/issues/2071

Do we also need to stop autorun at this point?

### DIFF
--- a/main/actions.js
+++ b/main/actions.js
@@ -1,10 +1,13 @@
 const { Ooniprobe } = require('./utils/ooni/ooniprobe')
+const { getAutorunHomeDir } = require('./utils/paths')
+const fs = require('fs-extra')
 const log = require('electron-log')
 const Sentry = require('@sentry/electron')
 
 const hardReset = () => {
   const ooni = new Ooniprobe()
   log.info('hardReset: performing a hard reset of the installation')
+  fs.removeSync(getAutorunHomeDir())
   return ooni.call(['reset', '--force'])
 }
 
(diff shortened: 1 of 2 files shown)


main/actions.js
```diff
@@ -1,10 +1,13 @@
 const { Ooniprobe } = require('./utils/ooni/ooniprobe')
+const { getAutorunHomeDir } = require('./utils/paths')
+const fs = require('fs-extra')
 const log = require('electron-log')
 const Sentry = require('@sentry/electron')
 
 const hardReset = () => {
   const ooni = new Ooniprobe()
   log.info('hardReset: performing a hard reset of the installation')
+  fs.removeSync(getAutorunHomeDir())
```
hellais, line 10: I think that what we ought to be doing is calling `ooniprobe reset --force` with the `OONI_HOME` environment variable set to `getAutorunHomeDir`.

That way, if there is any additional cleanup operation that needs to happen on hard reset, we can implement it inside of probe-cli directly and not have to remember to keep it in sync between the two directories.

hellais:
> Do we also need to stop autorun at this point?

I don't think we should do that. The autorun home directory will be re-created automatically the next time it runs, but it will be empty.

I guess one thing we need to think about is what is going to happen if the hard reset is executed while the autorun task is currently running.

bassosimone:
> I guess one thing we need to think about is what is going to happen if the hard reset is executed while the autorun task is currently running.

Yeah, this would be an interesting scenario

bassosimone (commented):
The diff looks good to me. I have spent some time thinking about what could happen if a hard reset occurs while an autorun is running, and I don't know what could happen. 🤷

PR #310, opened 2022-10-31
Release v3.8.1

hellais:
Includes fix for Signal test via probe-cli v3.16.4


hellais:
I fixed the tests I was able to fix easily, but wasn't able to reliably reproduce the failing ones and I suspect it's going to require more fighting with the CI, so I filed an issue to fix it later: https://github.com/ooni/probe/issues/2348.

